PR #11, opened 2019-05-27
Sdk performance fixes

DanielGuramulta:


sdk/src/frame_impl.cpp
```diff
@@ -60,8 +60,11 @@ aditof::Status FrameImpl::getData(aditof::FrameDataType dataType,
         float offset = m_details.cal_data.offset;
         float gain = m_details.cal_data.gain;
 
-        for (int i = 0; i < m_details.width * m_details.height / 2; i++) {
-            m_depthData[i] = (uint16_t)((float)m_depthData[i] * gain + offset);
+        if (gain != 1.0f || offset != 0.0f) {
```
dNechita: I'm not very excited about comparing floats. They are not always reliable. Maybe use an epsilon in these comparisons.

PR #21, opened 2019-05-29
sdk: windows-usb-device: check for nullptr

DanielGuramulta:


sdk/src/windows/usb_device_windows.cpp
```diff
@@ -750,6 +750,10 @@ aditof::Status UsbDevice::readEeprom(uint32_t address, uint8_t *data,
         }
     }
 
+    if (!pKsTopologyInfo) {
```
dNechita: After taking a closer look, this might not be required because I don't think that QueryInterface() can return success and, at the same time, set pKsTopologyInfo to nullptr.

PR #35, opened 2019-06-13
Add simple example

DanielGuramulta:


examples/CMakeLists.txt
```diff
@@ -2,3 +2,4 @@ cmake_minimum_required(VERSION 2.8)
 project(examples)
 
 add_subdirectory(aditof-demo)
+add_subdirectory(read-frame)
```
dNechita: Let's call this example: "first-frame"

PR #38, opened 2019-06-14
Add version to demo

DanielGuramulta:


examples/aditof-demo/main.cpp
```diff
@@ -9,9 +10,10 @@ int main(int argc, char **argv) {
     google::InitGoogleLogging(argv[0]);
     FLAGS_alsologtostderr = 1;
 
+    auto version = aditof::getApiVersion();
```
dNechita: use full type please. By reading this line, I have no idea what type version is.

PR #51, opened 2019-06-25
Apply calibration in device

DanielGuramulta:


sdk/src/dragonboard/local_device.cpp
```diff
@@ -671,3 +683,47 @@ aditof::Status LocalDevice::readLaserTemp(float &temperature) {
...
+
+    unsigned int width = m_implData->frameDetails.width;
+    unsigned int height = m_implData->frameDetails.height;
+
+    int16_t *cache = m_implData->calibration_cache[mode].cache;
+
+    unsigned int size = width * height / 2;
+
+    for (int i = 0; i < size; ++i) {
+        frame[i] = cache[frame[i]];
```
acozma: This will run faster if the frame pointer is incremented in the for loop and the statement becomes  *frame = cache[*frame];

sdk/src/linux/usb_device_linux.cpp
```diff
@@ -673,3 +689,47 @@ aditof::Status UsbDevice::readLaserTemp(float &temperature) {
 
     return Status::OK;
 }
+
+aditof::Status UsbDevice::setCalibrationParams(const std::string &mode,
```
dNechita: The implementations are very similar on all platforms. Group code that is duplicate into some utils functions or a helper object.

sdk/src/windows/usb_device_windows.cpp
```diff
@@ -1274,3 +1274,25 @@ aditof::Status UsbDevice::readLaserTemp(float &temperature) {
...
     return status;
 }
+
+aditof::Status UsbDevice::setCalibrationParams(const std::string & /*mode*/,
+                                               float /*gain*/,
+                                               float /*offset*/) {
+    using namespace aditof;
+    Status status = Status::OK;
+
+    // TO DO
```
dNechita: Will this be done in this PR?

sdk/src/linux/usb_device_linux.cpp
DanielGuramulta: Moved the two methods that compute and apply the calibration data in a Utils class.

sdk/src/utils.cpp
```diff
@@ -0,0 +1,39 @@
+#include "utils.h"
+
+uint16_t *aditof::Utils::getCalibrationData(float gain, float offset,
```
dNechita: Makes more sense to rename this to 'buildCalibrationCache()'

sdk/src/utils.cpp
```diff
@@ -0,0 +1,39 @@
...
+                                            int16_t maxPixelValue) {
+    uint16_t *cache = new uint16_t[maxPixelValue];
+    for (int16_t current = 0; current < maxPixelValue; ++current) {
+        cache[current] =
+            static_cast<int16_t>(static_cast<float>(current) * gain + offset);
+    }
+    return cache;
+}
+
+void aditof::Utils::applyCalibrationToFrame(uint16_t *calibrationData,
```
dNechita: 'calibrateFrame(frame, width, height, calibration_cache)'

sdk/src/utils.cpp
```diff
@@ -0,0 +1,39 @@
...
+        *(framePtr + 6) = *(cache + *(framePtr + 6));
+        *(framePtr + 7) = *(cache + *(framePtr + 7));
+    }
+
+    end += (size % 8);
+
+    for (; framePtr < end; framePtr++) {
+        *framePtr = *(cache + *framePtr);
+    }
+}
```
dNechita: No empty line

PR #53, opened 2019-07-02
Debian docker travis build

DanielGuramulta:


ci/travis/deps.sh
```diff
@@ -24,4 +24,21 @@ deps_deploy_doxygen() {
     echo_green "Doxygen version: " `doxygen --version`
 }
 
+deps_dragonboard() {
+    sudo apt-get -qq update
+	sudo service docker restart
+    sudo docker pull rycus86/arm64v8-debian-qemu
```
dNechita: Is 'sudo' required?

DanielGuramulta: The docker daemon is ran as root user, so all the client commands should use 'sudo'.

PR #63, opened 2019-08-14
Ci for ethernet impl

dNechita:


ci/travis/deps.sh
```diff
@@ -28,17 +43,7 @@ deps_dragonboard() {
...
-    [ -d "glog" ] || {
-       git clone https://github.com/google/glog
-    }
-    pushd glog
-    git checkout tags/v0.3.5
-    popd
-    popd
 }
 
+get_deps_source_code
```
DanielGuramulta: Shouldn't "get_deps_source_code" be called only in "deps_default()" and "deps_dragonboard()". There is no reason to get the source code for the dependencies when running builds that do: cppcheck, clang-format, doxygen.

dNechita: You are right. Thanks!

PR #75, opened 2019-08-28
build prefix incorrect

bitslip6:
path should be /opt not opt.  This prevents cmake from finding protobuf and fails the build.


acozma:
Thanks for contributing. This change has already been made, closing this PR.

PR #77, opened 2019-09-02
Camera range

DanielGuramulta:


DanielGuramulta:
> Looks good, the only required change is to make the near range 800 instead of 1000.
> Can merge into master once this is fixed.

I have fixed the near range. It should be 800 now!

PR #101, opened 2019-09-24
Device interface python bind

dNechita:
This implements #88

### DIFF
--- a/bindings/python/aditofpython.cpp
+++ b/bindings/python/aditofpython.cpp
@@ -1,4 +1,5 @@
 #include "pybind11/functional.h"
+#include "pybind11/numpy.h"
 #include "pybind11/pybind11.h"
 #include "pybind11/stl.h"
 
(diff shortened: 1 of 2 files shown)


bindings/python/aditofpython.cpp
```diff
@@ -140,5 +141,105 @@ PYBIND11_MODULE(aditofpython, m) {
...
+                 py::buffer_info buffInfo = data.request();
+                 uint8_t *ptr = static_cast<uint8_t *>(buffInfo.ptr);
+
+                 return device.writeEeprom(address, ptr, length);
+             },
+             py::arg("address"), py::arg("data"), py::arg("length"))
+        .def("readAfeRegisters",
+             [](DeviceInterface &device, py::array_t<uint16_t> address,
+                py::array_t<uint16_t> data, size_t length) {
+                 py::buffer_info addrBuffInfo = address.request(true);
```
grigorovici-andreea: change to be request(false) if not writable

bindings/python/aditofpython.cpp
```diff
@@ -140,5 +141,105 @@ PYBIND11_MODULE(aditofpython, m) {
...
+                 uint16_t *addrPtr = static_cast<uint16_t *>(addrBuffInfo.ptr);
+
+                 py::buffer_info dataBuffInfo = data.request(true);
+                 uint16_t *dataPtr = static_cast<uint16_t *>(dataBuffInfo.ptr);
+
+                 return device.readAfeRegisters(addrPtr, dataPtr, length);
+             },
+             py::arg("address"), py::arg("data"), py::arg("length"))
+        .def("writeAfeRegisters",
+             [](DeviceInterface &device, py::array_t<uint16_t> address,
```
grigorovici-andreea: same here (see above comm)

PR #102, opened 2019-09-25
Dnn python example

grigorovici-andreea:
This PR implements: #87.

- Added an object detection example in Python, using OpenCV DNN implementation. Combined Depth and IR images to get better results. 

- Changed the structure of the folder for better visibility of the examples.

- Added documentation for each example, and updated the general documentation for PyBindings.


bindings/python/readme.md
```diff
@@ -1,63 +1,20 @@
...
 
 | Directory/File | Description |
 | --------- | ----------- |
 | cmake | Contains the cmake files to find pybind11 |
 | pybind11 | Pybind11 header only library |
 | CMakeLists.txt | Rules to build the Python bindings |
-| first_frame.py | Python example to capture a frame |
+| first_frame | Python example to capture a frame |
+| dnn | Simple object detection example in Python |
+| device_example | Simple Python example of using Device Manager |
```
dNechita: Not sure where 'Device Manager' is coming from? It should say 'DeviceInterface' or 'device from the low-level API'

grigorovici-andreea: implemented with commit: e3a28ef

PR #103, opened 2019-09-25
sdk: network: Prevent from sending empty messages to server

### LINKED ISSUE #99: Fix ethernet interface lock on Windows

dNechita:
This should fix #99

### DIFF
--- a/sdk/src/network.cpp
+++ b/sdk/src/network.cpp
@@ -27,11 +27,16 @@ using namespace google::protobuf::io;
 using namespace payload;
 using namespace std;
 
+struct clientData {
+    bool hasFragments;
+    std::vector<char> data;
+};
+
 static struct lws_protocols protocols[] = {
     {
         "network-protocol",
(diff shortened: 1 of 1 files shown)


sdk/src/network.cpp
```diff
@@ -378,6 +397,8 @@ int Network::callback_function(struct lws *wsi,
         cout << endl << "Client is sending " << send_buff.func_name() << endl;
 #endif
         std::lock_guard<std::mutex> guard(m_mutex);
+        if (send_buff.DebugString().empty())
```
DanielGuramulta: Can't we use `ByteSizeLong()` instead of `DebugString().empty()` to check the size of the message?

dNechita: I tired it but doesn't work all the time. For some messages that have the DebugString() empty the ByteSizeLong() returns values > 0. I've made changes so that it checks whether the command (the function name) is valid or not. This should be enough. Thanks!

PR #110, opened 2019-10-01
scripts: linux: add setup script for linux

DanielGuramulta:
Addresses #82 / script for linux

### DIFF
--- a/scripts/linux/setup.sh
+++ b/scripts/linux/setup.sh
@@ -1,0 +1,192 @@
+#!/bin/bash
+
+print_help() {
+        echo "./setup [OPTIONS]"
+        echo ""
+        echo "-h|--help"
+        echo "        Print a usage message briefly summarizing the command line options available, then exit."
+        echo "-y|--yes"
+        echo "        Automatic yes to prompts."
+        echo "-b|--buildir"
+        echo "        Specify the build directory of the SDK."
+        echo "-d|--depsdir"
+        echo "        Specify the directory where the dependencies will be downloaded."
+        echo "-i|--depsinstalldir"
+        echo "        Specify the directory where the dependencies will be installed."
+        echo "-j|--jobs"
+        echo "        Specify the number of jobs to run in parallel when building dependencies and the SDK"
+        echo ""
+}
+
+install_required_packages() {
+        sudo apt-get -y install cmake libssl-dev libopencv-dev v4l-utils
+}
+
+get_deps_source_code() {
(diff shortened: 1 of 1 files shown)


scripts/linux/setup.sh
```diff
@@ -0,0 +1,180 @@
...
+        esac
+        done
+        set -- "${POSITIONAL[@]}" # restore positional parameters
+
+        if [[ "${display_help}" == "True" ]]; then
+                print_help
+                exit
+        fi
+        insta
+        ll_required_packages
```
dNechita: this seems broken in 2 lines

scripts/linux/setup.sh
```diff
@@ -0,0 +1,180 @@
...
+
+        get_deps_source_code ${deps_dir}
+
+        build_and_install_glog ${deps_dir}/glog ${deps_install_dir}/glog
+        build_and_install_protobuf ${deps_dir}/protobuf ${deps_install_dir}/protobuf
+        build_and_install_websockets ${deps_dir}/libwebsockets ${deps_install_dir}/websockets
+
+        pushd "${build_dir}"
+        cmake "${source_dir}" -DWITH_PYTHON=on -DWITH_OPENCV=on -DCMAKE_PREFIX_PATH="${deps_install_dir}/glog;${deps_install_dir}/protobuf;${deps_install_dir}/websockets"
+        make -j
```
dNechita: ${NUM_JOBS}?

scripts/linux/setup.sh
DanielGuramulta: Something must have happened when I formatted the file, Thank you!

scripts/linux/setup.sh
DanielGuramulta: I couldn't find a way to get the number of cores for the system. I think if it is not to much work an option like `-c x` for example that will set `NUM_JOBS=x` will work, or even using just `make` as it is right now will work. Indeed I forgot to remove the `-j`. Thanks.

dNechita: You can use `nproc --all` or use the -c option or both (use nproc by default and overwrite if user provides the core count)

DanielGuramulta: I have added a `-j|--jobs` option, as I see `-j` is used more often to specify the number of concurrent jobs. If that is not specified, the number of jobs will be the return value of `nproc --all`. Thanks for your suggestion.

PR #113, opened 2019-10-02
Multiple camera types support

dNechita:
Addresses #98


sdk/src/camera_factory.cpp
```diff
@@ -0,0 +1,21 @@
...
+#include <aditof/device_definitions.h>
+
+namespace aditof {
+
+Camera *CameraFactory::buildCamera(DeviceInterface *device) {
+    using namespace aditof;
+
+    DeviceDetails devDetails;
+    device->getDetails(devDetails);
+    switch (devDetails.sensorType) {
```
grigorovici-andreea: is this for future usage? as there is only 1 case, why switch?

sdk/src/target/usb_device.cpp
```diff
@@ -77,3 +77,8 @@ UsbDevice::applyCalibrationToFrame(uint16_t * /*frame*/,
                                    const std::string & /*mode*/) {
     return aditof::Status::GENERIC_ERROR;
 }
+
+aditof::Status
+UsbDevice::getDetails(aditof::DeviceDetails & /*details*/) const {
```
grigorovici-andreea: & /*details*/ 

Is this something I do not understand yet? or?

sdk/src/camera_factory.cpp
dNechita: Yes, it is for future usage to support at least another type of camera.

sdk/src/target/usb_device.cpp
dNechita: This is a stub implementation so most function parameters are not used. We comment them so we don't get warnings about unused parameter. At some point we will treat warnings as errors.

PR #116, opened 2019-10-02
bindings: opencv: aditof_opencv.h: add missing include guards

DanielGuramulta:


bindings/opencv/aditof_opencv.h
```diff
@@ -1,3 +1,6 @@
+#ifndef ADITOF_OPENCV
```
dNechita: We should try to be consistent with the guards naming: ADITOF_OPENCV_H

Issue #114, opened 2019-10-02
Main page documentation improvements


grigorovici-andreea:
Merge the _Examples Table_ and the _Bindings Examples Table_ into one big table with all the examples, with linking in the programming language Column.

PR #117, opened 2019-10-03
merged the 2 example tables

grigorovici-andreea:
Solves #114 .

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,35 +50,33 @@ For more details about the SDK check the links below:
   <tr>
     <td>aditof-demo</td>
     <td><a href="https://github.com/analogdevicesinc/aditof_sdk/tree/master/examples/aditof-demo"> C++ </a> </td>
-    <td>An application that uses OpenCV to display infrared and depth images</td>
+    <td>An application that displays the infrared and depth images</td>
   </tr>
   <tr>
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -41,19 +41,46 @@ For more details about the SDK check the links below:
...
+    <td>An application that uses OpenCV to display infrared and depth images</td>
+  </tr>
+  <tr>
+    <td rowspan="2">first-frame</td>
+    <td><a href="https://github.com/analogdevicesinc/aditof_sdk/tree/master/examples/first-frame"> C++ </a></td>
+    <td>An example code that shows the steps required to get to the point where camera frames can be captured</td>
+  </tr>
+  <tr>
+    <td><a href="https://github.com/analogdevicesinc/aditof_sdk/tree/master/bindings/python/examples/first_frame"> Python </a></td>
+    <td>Contains a python example to capture a frame</td>
```
dNechita: It might be better to have just one description for each example and that description should avoid mentioning any programming language since that's already covered in the language column. This way it will be easier to add more versions of the example for other languages.

grigorovici-andreea: Good point. did the necessary modifications. commit: 16a9d9a

PR #118, opened 2019-10-03
Cmake fixes

DanielGuramulta:


DanielGuramulta:
I have added a comment in the `cmake` readme about using `ldconfig` after installing.

PR #120, opened 2019-10-04
Dnn example improve

grigorovici-andreea:
implements: #108


bindings/opencv/dnn/main.cpp
```diff
@@ -122,48 +123,56 @@ int main(int argc, char *argv[]) {
...
             LOG(ERROR) << "Could not convert from frame to mat!";
             return 0;
         }
 
+        /* Obtain the ir mat from the frame */
+        cv::Mat irMat;
+        status = fromFrameToIrMat(frame, irMat);
+        if (status != Status::OK) {
+            LOG(ERROR) << "Could not convert from frame to mat!";
+            return 0;
```
dNechita: The 0 return value is associated with success. You could return something different from 0, like -1.

bindings/opencv/dnn/readme.md
```diff
@@ -1,6 +1,6 @@
 # DNN example
 
 ### Overview
-This example demonstrates object detection on the depth frame using the [SSD Mobilenet object detection example from OpenCV](https://github.com/opencv/opencv/blob/3.4.0/samples/dnn/ssd_mobilenet_object_detection.cpp) and the Aditof SDK.
+This example demonstrates object detection on the a combination between depth and ir frames using the [SSD Mobilenet object detection example from OpenCV](https://github.com/opencv/opencv/blob/3.4.0/samples/dnn/ssd_mobilenet_object_detection.cpp) and the Aditof SDK.
```
dNechita: I think 'a' or 'the' is extra here because the hole sentence doesn't sound right.

bindings/opencv/dnn/main.cpp
grigorovici-andreea: modified in commit: 12ef2f1c34a3c0a54f11b489560750b1f119899f

bindings/opencv/dnn/readme.md
grigorovici-andreea: modified in commit: 12ef2f1c34a3c0a54f11b489560750b1f119899f

PR #125, opened 2019-10-10
sdk: clamp the values of each pixel after calibration to the range

DanielGuramulta:

### DIFF
--- a/sdk/src/linux/usb_device_linux.cpp
+++ b/sdk/src/linux/usb_device_linux.cpp
@@ -635,15 +635,16 @@ aditof::Status UsbDevice::readLaserTemp(float &temperature) {
 }
 
 aditof::Status UsbDevice::setCalibrationParams(const std::string &mode,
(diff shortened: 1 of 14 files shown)


sdk/src/linux/usb_device_linux.cpp
```diff
@@ -657,10 +657,6 @@ aditof::Status UsbDevice::applyCalibrationToFrame(uint16_t *frame,
     float gain = m_implData->calibration_cache[mode].gain;
     float offset = m_implData->calibration_cache[mode].offset;
 
-    if (equal(gain, 1.0f) && equal(offset, 0.0f)) {
```
dNechita: What about the other implementations (ethernet, local, usb-windows)?

DanielGuramulta: My bad. Forgot that this is implemented in each device. Thanks!

PR #130, opened 2019-10-14
examples: add first frame examples using ip

DanielGuramulta:


examples/first-frame-ethernet/readme.md
```diff
@@ -0,0 +1,6 @@
+# first-frame-ethernet sample code
```
dNechita: Can you add an example of how to run this example?

PR #146, opened 2019-10-25
Aditofdemo improvements

grigorovici-andreea:
solves #92

### DIFF
--- a/examples/aditof-demo/aditofdemoview.cpp
+++ b/examples/aditof-demo/aditofdemoview.cpp
@@ -196,7 +196,7 @@ void AdiTofDemoView::render() {
             playbackChecked = xorValue & 1;
         }
 
-        // play button group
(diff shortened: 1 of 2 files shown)


examples/aditof-demo/aditofdemoview.cpp
```diff
@@ -190,6 +196,15 @@ void AdiTofDemoView::render() {
             playbackChecked = xorValue & 1;
         }
 
+        // play button group
+        int btnGroupView = separatedViewChecked << 1 | blendedViewChecked;
```
dNechita: You could add a description on what exactly is being done here. For e.g.: 'Creating a group of buttons where only one can be active at a time.'

grigorovici-andreea: solved with: e7cff9e59a801ac79ecda9ecbc7053883bd30962

PR #157, opened 2019-11-11
sdk: Add a calibration mode

dNechita:
In calibration mode, client code provides the firmware.
Also the data is left untouched, no corrections or clamping is done on the frames.


dNechita:
> I'd suggest calling the new mode "custom" instead of "calibration". In the end this is the mode where the user loads a custom firmware into the AFE

Done.

PR #158, opened 2019-11-13
Code improvements

dNechita:


apps/uvc-app/uvc-gadget.cpp
```diff
@@ -1652,7 +1647,7 @@ static int uvc_events_process_data(struct uvc_device *dev,
              */
             return -EINVAL;
 #if 0
-          if (*val > PU_BRIGHTNESS_MAX_VAL) {
+          if (*(unsigned int *)data->data > PU_BRIGHTNESS_MAX_VAL) {
```
DanielGuramulta: Is there any reason to not use a c++ style cast here instead of the c style cast?

dNechita: it's in a #if 0 block and I don't think we will be using it anytime soon. So I didn't bother adding a static cast.

DanielGuramulta: Did not see the #if 0. Makes sens then.

PR #161, opened 2019-11-18
Matlab bindings

DanielGuramulta:


DanielGuramulta:
The cmake implementation uses this https://github.com/analogdevicesinc/aditof_sdk/blob/dfb03c8ab383cc5ed7c3ed88d0a08d1c88f0f902/cmake/FindMatlabWithImageAcquisitionToolbox.cmake cmake file which calls find_package for matlab. On a computer with a fresh install of matlab it had trouble finding the root directory (still investigating why). In that case providing the path fixed the issue. 

test_camera.m and test_load.m will be updated. This is why I made this a draft pull request.
I will have a look into building some tests for our adaptor asap.

We also have on TravisCI a build that runs `cppcheck` against our code. I will make sure it also scans the source code of our adaptor.

PR #164, opened 2019-11-27
Add smart ptrs

DanielGuramulta:
This pull request adds back smart pointers in the SDK.

### DIFF
--- a/apps/uvc-app/uvc-gadget.cpp
+++ b/apps/uvc-app/uvc-gadget.cpp
@@ -284,7 +284,8 @@ void stopHandler(int code) { stop.store(true); }
  * V4L2 streaming related
  */
 
-static int v4l2_process_data(LocalDevice *device, uvc_device *udev) {
+static int v4l2_process_data(std::shared_ptr<LocalDevice> device,
+                             uvc_device *udev) {
     int ret;
     struct v4l2_buffer ubuf;
     struct v4l2_buffer vbuf;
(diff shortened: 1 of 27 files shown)


apps/uvc-app/uvc-gadget.cpp
```diff
@@ -284,7 +284,8 @@ void stopHandler(int code) { stop.store(true); }
  * V4L2 streaming related
  */
 
-static int v4l2_process_data(LocalDevice *device, uvc_device *udev) {
+static int v4l2_process_data(std::shared_ptr<LocalDevice> device,
```
grigorovici-andreea: uvc_device *udev 
Astea de ce nu vor fi schimbate?

sdk/src/frame.cpp
```diff
@@ -5,27 +5,17 @@ namespace aditof {
...
-    if (m_impl)
-        delete m_impl;
-}
+Frame::~Frame() = default;
 
-Frame::Frame(Frame &&other) noexcept {
-    this->m_impl = other.m_impl;
-    other.m_impl = nullptr;
-};
+Frame::Frame(Frame &&other) noexcept = default;
```
grigorovici-andreea: Oare nu ar fi mai ok sa fie ca aici, fara nume (other) : System::System(System &&)

mhennerich: Friendly reminder - Interaction on github should be in plain English please

apps/uvc-app/uvc-gadget.cpp
DanielGuramulta: The uvc gadget was initially implemented using C, the main purpose of this PR is to add smart pointers to the SDK api.

sdk/src/frame.cpp
DanielGuramulta: Modified to Frame::Frame(Frame &&). Thanks.

PR #168, opened 2019-12-02
Demo connection options

grigorovici-andreea:
examples/aditof_demo: 
Added options for connection mode:
 - Depending on the platform used, the user can switch between the available connection modes.
Fixed small bug in the UI for "Small-signal removal"

### DIFF
--- a/examples/aditof-demo/CMakeLists.txt
+++ b/examples/aditof-demo/CMakeLists.txt
@@ -16,6 +16,14 @@ else()
 	endif()
 endif()
 
+if( WIN32 OR UNIX )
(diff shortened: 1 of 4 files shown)


examples/aditof-demo/CMakeLists.txt
```diff
@@ -16,6 +16,14 @@ else()
 	endif()
 endif()
 
+if( WIN32 OR UNIX )
```
dNechita: Rethink the order. Dragonboard and Raspberry PI are Unix as well.

grigorovici-andreea: done

PR #173, opened 2019-12-16
Fix wrong displayed color

DanielGuramulta:
Fixes bug described in issue: #171

### DIFF
--- a/sdk/src/target/local_device.cpp
+++ b/sdk/src/target/local_device.cpp
@@ -394,8 +394,10 @@ aditof::Status LocalDevice::program(const uint8_t *firmware, size_t size) {
     size_t readBytes = 0;
 
     if (size <= CTRL_PACKET_SIZE) {
+        memset(buf + size, 0, CTRL_PACKET_SIZE);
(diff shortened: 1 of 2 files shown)


sdk/src/target/local_device.cpp
```diff
@@ -394,8 +394,10 @@ aditof::Status LocalDevice::program(const uint8_t *firmware, size_t size) {
     size_t readBytes = 0;
 
     if (size <= CTRL_PACKET_SIZE) {
+        memset(buf, 0, CTRL_PACKET_SIZE);
+        memcpy(buf, firmware, size);
```
dNechita: Why not set to 0 the remaining bytes?

DanielGuramulta: Fixed the issue. Thanks for noticing!

DanielGuramulta:
From what I tested on Host over USB on Linux, and @dNechita on Windows, it seems that: #144 is also fixed by this PR. The demo application didn't crash anymore when playing with the small signal feature.

PR #177, opened 2019-12-18
Matlab doc

DanielGuramulta:


bindings/matlab/test_camera_ethernet.m
```diff
@@ -0,0 +1,10 @@
+
+%% The second device will always be a generic ethernet device where the ip
```
dNechita: .. where the ip specified .. The ip "is" specified. Seems that you're missing a verb or something

DanielGuramulta: Fixed. Thanks for noticing!

PR #179, opened 2019-12-18
Expose noise reduction in sdk

dNechita:
Implements #136


bindings/python/aditofpython.cpp
```diff
@@ -4,6 +4,7 @@
 #include "pybind11/stl.h"
 
 #include <aditof/aditof.h>
+#include <aditof/camera_96tof1_specifics.h>
```
DanielGuramulta: Does it make sense to also add this include in `aditof/aditof.h`?

dNechita: No. There are use cases where only the Camera interface is enough. Applications that require the extended interface will have to include the extra header which does not necessarily need to be camera_96tof1_specifics.h. It can be a header with specifics for another camera.

DanielGuramulta: Ok, that makes sense.

Issue #176, opened 2019-12-18
Python bindings: Near mode correction


grigorovici-andreea:
Investigate the source of extra noise on the **near mode** in python bindings examples.

grigorovici-andreea:
As IR is not calibrated => values that go over 'range' are ruining the IR image. 

We need to figure if it is worth to do this clamping over IR.

grigorovici-andreea:
Looks like the problem comes from the wrong usage of range in building the IR image. Instead of the range, we will use the max possible value of a pixel. Because there are only 12 bits that are usable from 16, the maximal value = 4065.

PR #191, opened 2020-01-14
Changed the calculation of the IR frame in all examples.

grigorovici-andreea:
Instead of the range, we will use the max possible value of a pixel.
Because there are only 12 bits that are usable from 16, the maximal value = 4095.

Fixes: #176.

### DIFF
--- a/bindings/matlab/source_adaptor.cpp
+++ b/bindings/matlab/source_adaptor.cpp
@@ -284,6 +284,18 @@ int SourceAdaptor::getCurrentHwRange() {
     return cameraDetails.range;
 }
 
+int SourceAdaptor::getCurrentBitCount() {
+    aditof::CameraDetails cameraDetails;
+
+    if (!m_camera) {
+        return 0;
+    }
+
+    m_camera->getDetails(cameraDetails);
+
+    return cameraDetails.bitCount;
+}
+
 std::shared_ptr<aditof::Frame> SourceAdaptor::getFrameFromHwDevice() {
     if (!m_camera) {
         return nullptr;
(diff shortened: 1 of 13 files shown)


DanielGuramulta:
the maximum value = `4095`

bindings/matlab/source_adaptor.cpp
```diff
@@ -509,17 +521,17 @@ void SourceAdaptor::sendFrame(SourceAdaptor *adaptor) {
 
         auto frame = adaptor->getFrameFromHwDevice();
         int currentRange = adaptor->getCurrentHwRange();
-
+		int currentMaxPixelValue =  (1 << adaptor->getCurrentBitCount()) - 1;
```
DanielGuramulta: Extra tab / spaces

bindings/matlab/source_adaptor.h
```diff
@@ -177,6 +177,8 @@ class SourceAdaptor : public imaqkit::IAdaptor {
     void setMode(int16_t mode);
 
     int getCurrentHwRange();
+	
+	int getCurrentBitCount();
```
DanielGuramulta: Extra spaces

sdk/include/aditof/camera_definitions.h
```diff
@@ -59,6 +59,11 @@ struct CameraDetails {
      * @brief The range of the camera in mm for the operating mode
      */
     int range;
+
+    /**
+     * @brief The number of bits used for ADC
```
dNechita: On a second thought, maybe not mention the ADC. Sorry.
What about: "The number of bits used for representing one pixel data." ?

grigorovici-andreea: solved. thanks

bindings/matlab/source_adaptor.h
grigorovici-andreea: solved. thanks

bindings/matlab/source_adaptor.cpp
grigorovici-andreea: solved. thanks

Issue #185, opened 2019-12-20
Demo: why use a reduced color map for medium and far modes


dNechita:
The depth is mapped to colors, spanning from red to magenta.
In medium and far modes, the minimum length that can be measured is higher than in near mode but for these minimum values we don't start mapping from red.
So, the color map should be adjusted to the measuring interval. The red should correspond to the lowest measurement value no matter which mode is enabled.

grigorovici-andreea:
The minum range of the system:
              Theory                                                  Real
Near - 250 mm                                        ||     0 mm
Medium - 300 mm (RevB: 800 mm)        ||     390 mm
Far - 3000 mm                                         ||     2000 mm

PR #195, opened 2020-01-20
Demo: Colored depth map values correction by using minimum range

grigorovici-andreea:
Added minimum extremity range in camera details.
Used the minimum extremity range to correct the values in colored depth map.

Solves #185 

### DIFF
--- a/bindings/python/aditofpython.cpp
+++ b/bindings/python/aditofpython.cpp
@@ -53,7 +53,8 @@ PYBIND11_MODULE(aditofpython, m) {
         .def_readwrite("mode", &aditof::CameraDetails::mode)
         .def_readwrite("frameType", &aditof::CameraDetails::frameType)
         .def_readwrite("connection", &aditof::CameraDetails::connection)
-        .def_readwrite("range", &aditof::CameraDetails::range)
+        .def_readwrite("minDepth", &aditof::CameraDetails::minDepth)
+        .def_readwrite("maxDepth", &aditof::CameraDetails::maxDepth)
         .def_readwrite("bitCount", &aditof::CameraDetails::bitCount);
 
     // Helpers
--- a/sdk/include/aditof/camera_definitions.h
+++ b/sdk/include/aditof/camera_definitions.h
@@ -56,9 +56,16 @@ struct CameraDetails {
     ConnectionType connection;
 
     /**
-     * @brief The range of the camera in mm for the operating mode
+     * @brief The maximum distance (in millimeters) the camera can measure in
+     * the current operating mode.
      */
-    int range;
+    int maxDepth;
+
+    /**
+     * @brief The minimum distance (in millimeters) the camera can measure in
+     * the current operating mode.
+     */
+    int minDepth;
 
     /**
      * @brief The number of bits used for representing one pixel data.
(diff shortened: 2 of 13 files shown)


sdk/include/aditof/camera_definitions.h
```diff
@@ -56,9 +56,16 @@ struct CameraDetails {
     ConnectionType connection;
 
     /**
-     * @brief The range of the camera in mm for the operating mode
+     * @brief The maximum extremity of the range of the camera in mm for the
+     * operating mode
      */
-    int range;
+    int rangeMax;
```
dNechita: how about maxDepth and minDepth? It's more descriptive, makes it clear that this is for depth and not IR.

sdk/include/aditof/camera_definitions.h
```diff
@@ -56,9 +56,16 @@ struct CameraDetails {
     ConnectionType connection;
 
     /**
-     * @brief The range of the camera in mm for the operating mode
+     * @brief The maximum extremity of the range of the camera in mm for the
```
dNechita: The maximum depth the camera can measure in the current operating mode. Distance is in millimeters.

sdk/src/camera_96tof1.cpp
```diff
@@ -61,9 +61,12 @@ aditof::Status Camera96Tof1::setMode(const std::string &mode,
 
     LOG(INFO) << "Chosen mode: " << mode.c_str();
 
-    std::vector<std::pair<std::string, int>> modeRanges = {
+    std::vector<std::pair<std::string, int>> modeRangesMax = {
```
dNechita: You could create a struct which can describe a range. Make it internal to this class. And have just one array. You won't need to use the second find_if either.

sdk/include/aditof/camera_definitions.h
grigorovici-andreea: done, thanks

sdk/include/aditof/camera_definitions.h
grigorovici-andreea: done, thanks

sdk/src/camera_96tof1.cpp
grigorovici-andreea: done, thanks

bindings/python/aditofpython.cpp
```diff
@@ -53,7 +53,7 @@ PYBIND11_MODULE(aditofpython, m) {
         .def_readwrite("mode", &aditof::CameraDetails::mode)
         .def_readwrite("frameType", &aditof::CameraDetails::frameType)
         .def_readwrite("connection", &aditof::CameraDetails::connection)
-        .def_readwrite("range", &aditof::CameraDetails::range)
+        .def_readwrite("range", &aditof::CameraDetails::rangeMax)
```
dNechita: This line should be: .def_readwrite("maxDepth", &aditof::CameraDetails::maxDepth). There's no rangeMax anymore. The python bindings shouldn't even compile anymore.
Also, can you create another binding for minDepth as well?

grigorovici-andreea: totally forgot about bindings. working on it now.

grigorovici-andreea: changed. thank you.

Issue #196, opened 2020-01-21
Update the instructions on how to unpack the SD card image


acozma:
The latest SD card image archive was compressed using xz. The unpack instructions [here](https://github.com/analogdevicesinc/aditof_sdk/blob/master/doc/sdcard_burn.md) must be updated since this compression method will be used from now on.

Also, need to add .img extension to the file that's in the xz archive.

dNechita:
The SD card images have been updated. Closing this.

PR #197, opened 2020-01-21
doc: Update instructions for writing an SD card

dNechita:
Addresses #196

### DIFF
--- a/doc/sdcard_burn.md
+++ b/doc/sdcard_burn.md
@@ -8,9 +8,9 @@ Writing the image on the SD card will take about 15 to 20 minutes for a 16GB ima
     ~~~sh
     sudo apt-get install xz-utils
     ~~~
-* Unpack the .xz archive
+* Unpack the .tar.xz archive
     ~~~sh
-    unxz name-of-the-image.xz
+    tar -xf name-of-the-image.tar.xz
     ~~~
 * Write the image on the SD card
     ~~~sh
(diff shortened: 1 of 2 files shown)


doc/sdcard_burn.md
```diff
@@ -1,26 +1,34 @@
...
     ~~~
 * Write the image on the SD card
     ~~~sh
-    dd if=xxx.img of=/dev/xxx bs=4M
+    dd if=name-of-the-image.img of=/dev/xxx bs=4M
     sync
     ~~~
     
-Note: The **lsblk** command can be used to help identifying the device to be specified for the argument: **of=/dev/xxx**.
+Note 1: The following command can be used to help identifying the device than needs to be specified for the argument **of=/dev/xxx**:
```
grigorovici-andreea: I think here should be **that**: "....the device than needs to be specified...."

doc/sdcard_burn.md
```diff
@@ -1,26 +1,34 @@
...
+~~~sh
+    lsblk
+~~~
+Note 2: It is very likely that the **dd** command will have to be executed with **sudo** priviledges:
```
grigorovici-andreea: here is: **privileges**

doc/sdcard_burn.md
dNechita: yes, fixed. Thanks!

doc/sdcard_burn.md
dNechita: it is. Fixed.

PR #199, opened 2020-01-22
Calibration remove old lib

bhatiah:


dNechita:
There's one more import of old lib left in tof_calib/intrinsic_calibration.py
Can you please squash everything in one commit and then sign-off the commit?

Other than that, it looks good. Thanks!

PR #201, opened 2020-01-23
sdk: Add option to select the revision in camera specifics

grigorovici-andreea:
Revision can be selected and the value for the MaxDepth changes accordingly.
Added example of usage in the demo application.

Solves #186.

### DIFF
--- a/examples/aditof-demo/aditofdemocontroller.h
+++ b/examples/aditof-demo/aditofdemocontroller.h
@@ -58,6 +58,8 @@ class AdiTofDemoController {
     aditof::Status enableNoiseReduction(bool en);
     aditof::Status setNoiseReductionThreshold(uint16_t threshold);
 
+    aditof::Status setCameraRevision(const std::string &revision);
+
   private:
     void captureFrames();
 
--- a/sdk/include/aditof/camera_96tof1_specifics.h
+++ b/sdk/include/aditof/camera_96tof1_specifics.h
@@ -11,6 +11,12 @@ class Camera96Tof1;
 namespace aditof {
 
 /**
+ * @enum Revision
+ * @brief Enumerates camera revisions
+ */
+enum class Revision { RevB, RevC };
+
+/**
  * @class Camera96Tof1Specifics
  * @brief Implements the extened API that is specific for the 96 TOF1 camera.
  */
@@ -48,6 +54,18 @@ class SDK_API Camera96Tof1Specifics : public CameraSpecifics {
      */
     uint16_t noiseReductionThreshold() const;
 
+    /**
+     * @brief Sets the camera revision type.
+     * @return Status
+     */
+    Status setCameraRevision(Revision revision);
+
+    /**
+     * @brief Returns the current camera revision type.
+     * @return Revision
+     */
+    Revision getRevision() const;
+
   private:
     Status setTresholdAndEnable(uint16_t treshold, bool en);
 
@@ -55,6 +73,7 @@ class SDK_API Camera96Tof1Specifics : public CameraSpecifics {
     Camera96Tof1 *m_camera;
     bool m_noiseReductionOn;
     uint16_t m_noiseReductionThreshold;
+    Revision m_revision;
 };
 
 } // namespace aditof
--- a/sdk/src/camera_96tof1.cpp
+++ b/sdk/src/camera_96tof1.cpp
@@ -6,16 +6,38 @@
 #include <aditof/frame_operations.h>
 
 #include <algorithm>
+#include <array>
 #include <fstream>
 #include <glog/logging.h>
 #include <iterator>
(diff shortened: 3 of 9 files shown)


examples/aditof-demo/aditofdemocontroller.h
```diff
@@ -58,6 +58,8 @@ class AdiTofDemoController {
     aditof::Status enableNoiseReduction(bool en);
     aditof::Status setNoiseReductionThreshold(uint16_t threshold);
 
+    aditof::Status setCameraRevision(std::string revision);
```
dNechita: It is a good practice to pass argument by reference. Another good practice: argument should be const as the intention is not to modify it.

sdk/src/camera_96tof1.cpp
```diff
@@ -15,7 +15,13 @@ static const std::string skCustomMode = "custom";
 Camera96Tof1::Camera96Tof1(std::unique_ptr<aditof::DeviceInterface> device)
     : m_specifics(std::make_shared<aditof::Camera96Tof1Specifics>(
           aditof::Camera96Tof1Specifics(this))),
-      m_device(std::move(device)), m_devStarted(false) {}
+      m_device(std::move(device)), m_devStarted(false) {
+
+    // initialize range values with the default data for revision C
```
dNechita: We assume the default is revision C in two places: here and in Camera96Tof1Specifics. Let's just do it once. Interrogate the revision in Camera96Tof1Specifics and use it here.
Also it would be nice to create a table with range values for each revision. Makes things clearer and you won't even need the setRangeValues() anymore.

dNechita:
Since these changes extend the API, python bindings should be updated as well!

examples/aditof-demo/aditofdemocontroller.h
grigorovici-andreea: "It is a good practice to pass argument by reference" I thought this applies only when the argument is big like a structure that occupies a lot of memory. I will change. thanks

dNechita: String can be one of those big structures

sdk/include/aditof/camera_96tof1_specifics.h
```diff
@@ -10,6 +10,12 @@ class Camera96Tof1;
 
 namespace aditof {
 
+/**
+ * @enum RevisionType
+ * @brief Provides the types of revision of the Camera
+ */
+enum class RevisionType { RevA, RevB, RevC };
```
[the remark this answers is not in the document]
grigorovici-andreea: changed, thanks

sdk/include/aditof/camera_96tof1_specifics.h
```diff
@@ -48,13 +54,26 @@ class SDK_API Camera96Tof1Specifics : public CameraSpecifics {
      */
     uint16_t noiseReductionThreshold() const;
 
+    /**
+     * @brief Sets the camera revision type.
+     * @return Status
+     */
+    Status setCameraRevision(std::string revision);
```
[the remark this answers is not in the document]
grigorovici-andreea: changed, thanks

examples/aditof-demo/aditofdemocontroller.h
grigorovici-andreea: changed, thanks

sdk/src/camera_96tof1.cpp
grigorovici-andreea: changed, thanks

sdk/src/camera_96tof1.cpp
```diff
@@ -59,11 +66,13 @@ aditof::Status Camera96Tof1::setMode(const std::string &mode,
...
 
-    LOG(INFO) << "Chosen mode: " << mode.c_str();
-
-    std::vector<struct rangeStruct> rangeValues = {
-        {"near", 250, 800}, {"medium", 300, 4500}, {"far", 3000, 6000}};
+    // Set the values specific to the Revision requested
+    auto cam96tof1Specifics =
+        std::dynamic_pointer_cast<Camera96Tof1Specifics>(m_specifics);
+    Revision revision = cam96tof1Specifics->getRevision();
+    m_rangeValues = RangeValuesForRevision.at(static_cast<int>(revision));
```
DanielGuramulta: Maybe use a map here to store the range values for revision

grigorovici-andreea: done. thanks

examples/aditof-demo/aditofdemocontroller.h
dNechita: nothing has changed. Looks the same.

grigorovici-andreea: sorry, now there should be the right modification

PR #204, opened 2020-02-04
sdk: Fix buffer overrun and memory leak in device enumerator

dNechita:
Bug is in the dragonboard implementation.


dNechita:
> Looks good, but CI failed due to formatting.

Will fix, thanks!

PR #213, opened 2020-02-17
bindings: ros: add initial version of ROS package

andreeasandulescu:
Added bindings for ros PointCloud2 message type

### DIFF
--- a/bindings/ros/aditof_roscpp/include/aditof_roscpp/aditof_sensor_msg.h
+++ b/bindings/ros/aditof_roscpp/include/aditof_roscpp/aditof_sensor_msg.h
@@ -1,0 +1,15 @@
+#ifndef ADITOF_SENSOR_MSG_H
+#define ADITOF_SENSOR_MSG_H
+
+#include <aditof/camera.h>
+#include <ros/publisher.h>
+
+class AditofSensorMsg {
+  public:
+    virtual ~AditofSensorMsg() = default;
+    virtual void
+    FrameDataToMsg(const std::shared_ptr<aditof::Camera> &camera) = 0;
+    virtual void publishMsg(const ros::Publisher &pub) = 0;
+};
+
+#endif // ADITOF_SENSOR_MSG_H
--- a/bindings/ros/aditof_roscpp/include/aditof_roscpp/aditof_utils.h
+++ b/bindings/ros/aditof_roscpp/include/aditof_roscpp/aditof_utils.h
@@ -1,0 +1,18 @@
+#ifndef ADITOF_UTILS_H
+#define ADITOF_UTILS_H
+
+#include <aditof/camera.h>
+
+std::shared_ptr<aditof::Camera> initCameraEthernet(int argc, char **argv);
+
+void applyNoiseReduction(const std::shared_ptr<aditof::Camera> &camera,
+                         int argc, char **argv);
+uint16_t *getNewFrame(const std::shared_ptr<aditof::Camera> &camera,
+                      aditof::Frame *frame);
+
+aditof::IntrinsicParameters
+getIntrinsics(const std::shared_ptr<aditof::Camera> &camera);
+int getRangeMax(const std::shared_ptr<aditof::Camera> &camera);
+int getRangeMin(const std::shared_ptr<aditof::Camera> &camera);
+
+#endif // ADITOF_UTILS_H
--- a/bindings/ros/aditof_roscpp/include/aditof_roscpp/message_factory.h
+++ b/bindings/ros/aditof_roscpp/include/aditof_roscpp/message_factory.h
@@ -1,0 +1,18 @@
+#ifndef MESSAGE_FACTORY_H
+#define MESSAGE_FACTORY_H
+
+#include "pointcloud2_msg.h"
+
+enum class MessageType {
+    sensor_msgs_PointCloud2
+    // TODO: add support for other types of messages
+    //,sensor_msgs_Image, sensor_msgs_CameraInfo
+};
+
+class MessageFactory {
+  public:
+    static AditofSensorMsg *
+    create(const std::shared_ptr<aditof::Camera> &camera, MessageType type);
+};
+
+#endif // MESSAGE_FACTORY_H
--- a/bindings/ros/aditof_roscpp/src/aditof_utils.cpp
+++ b/bindings/ros/aditof_roscpp/src/aditof_utils.cpp
@@ -1,0 +1,131 @@
+#include "aditof_utils.h"
+
+#include <aditof/camera_96tof1_specifics.h>
+#include <aditof/frame.h>
+#include <aditof/system.h>
+#include <glog/logging.h>
+#include <ros/ros.h>
+
+using namespace aditof;
+std::shared_ptr<Camera> initCameraEthernet(int argc, char **argv) {
+    google::InitGoogleLogging(argv[0]);
+    FLAGS_alsologtostderr = 1;
+    Status status = Status::OK;
+
+    if (argc < 2) {
+        LOG(ERROR) << "No ip provided!";
+        return nullptr;
+    }
+
+    std::string ip = argv[1];
+
+    System system;
+    status = system.initialize();
+    if (status != Status::OK) {
+        LOG(ERROR) << "Could not initialize system!";
+        return nullptr;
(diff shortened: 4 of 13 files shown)


bindings/ros/aditof_roscpp/include/aditof_roscpp/aditof_utils.h
```diff
@@ -0,0 +1,22 @@
+#ifndef ADITOF_UTILS_H
+#define ADITOF_UTILS_H
+
+#include <aditof/camera.h>
+#include <aditof/camera_96tof1_specifics.h>
```
dNechita: This is no required here, in this header. Also not system, glog and probably not ros.

bindings/ros/aditof_roscpp/src/aditof_utils.cpp
```diff
@@ -0,0 +1,125 @@
...
+
+using namespace aditof;
+std::shared_ptr<Camera> initCameraEthernet(int argc, char **argv) {
+    google::InitGoogleLogging(argv[0]);
+    FLAGS_alsologtostderr = 1;
+    Status status = Status::OK;
+
+    if (argc < 2) {
+        LOG(ERROR) << "No ip provided!";
+        return std::shared_ptr<Camera>(nullptr);
```
dNechita: why not return nullptr?

bindings/ros/aditof_roscpp/src/aditof_utils.cpp
```diff
@@ -0,0 +1,125 @@
...
+    status = camera->initialize();
+    if (status != Status::OK) {
+        LOG(ERROR) << "Could not initialize camera!";
+        return std::shared_ptr<Camera>(nullptr);
+    }
+
+    std::vector<std::string> frameTypes;
+    camera->getAvailableFrameTypes(frameTypes);
+    if (frameTypes.empty()) {
+        ROS_INFO("no frame type avaialble!\n");
```
dNechita: Why info and not error?

bindings/ros/aditof_roscpp/include/aditof_roscpp/generic_msg.h
```diff
@@ -0,0 +1,11 @@
+#ifndef GENERIC_MSG_H
+#define GENERIC_MSG_H
+
+#include <aditof/camera.h>
+
+class GenericMsg {
```
dNechita: Since classes will inherit from this one, this class must have a virtual destructor!

dNechita: Just naming it 'Message' is generic enough.
Since it is a message from aditof sensor why not name this class accordingly?

bindings/ros/aditof_roscpp/include/aditof_roscpp/message_factory.h
```diff
@@ -0,0 +1,15 @@
+#ifndef MESSAGE_FACTORY_H
+#define MESSAGE_FACTORY_H
+
+#include "pointcloud2_msg.h"
+
+enum MessageType {
```
dNechita: in C++ lets use enum class which is a more strong type.

bindings/ros/aditof_roscpp/src/pointcloud2_msg.cpp
```diff
@@ -0,0 +1,103 @@
...
+            *iter_x = x_n;
+            *iter_y = y_n;
+            *iter_z = z_n;
+
+            *iter_rgb = frameData[i * msg.width + j];
+        }
+    }
+}
+
+void PointCloud2Msg::createMessage(const std::shared_ptr<Camera> &camera)
```
dNechita: This looks like a method that fills the message not create it. The factory creates the message.

bindings/ros/aditof_roscpp/include/aditof_roscpp/message_factory.h
```diff
@@ -0,0 +1,15 @@
...
+    sensor_msgs_PointCloud2, sensor_msgs_Image, sensor_msgs_CameraInfo
+};
+
+class MessageFactory {
+public:
+    static aditof::Status create(const std::shared_ptr<aditof::Camera> &camera, MessageType type, GenericMsg& msg);
```
dNechita: A typical factory method returns the newly constructed object not a status.

bindings/ros/aditof_roscpp/src/message_factory.cpp
```diff
@@ -0,0 +1,20 @@
+#include "message_factory.h"
+
+aditof::Status MessageFactory::create(const std::shared_ptr<aditof::Camera> &camera, MessageType type, GenericMsg& msg)
+{
+    switch(type)
+    {
+        case sensor_msgs_PointCloud2:
+            msg.createMessage(camera);
+            return aditof::Status::OK;
+        case sensor_msgs_Image:
```
dNechita: if sensor_msgs_Image and others are not yet supported, don't include them here and in the enum. Otherwise clients could try to use them and discover at run-time that they are not supported. Just comment them in the enum and mark them as to-do.

bindings/ros/aditof_roscpp/include/aditof_roscpp/aditof_sensor_msg.h
```diff
@@ -0,0 +1,17 @@
+#ifndef GENERIC_MSG_H
```
dNechita: include guards no longer match filename and class name

bindings/ros/aditof_roscpp/include/aditof_roscpp/aditof_sensor_msg.h
```diff
@@ -0,0 +1,17 @@
...
+
+class AditofSensorMsg {
+  public:
+    virtual ~AditofSensorMsg() = 0;
+    virtual void
+    convertFrameToMsg(const std::shared_ptr<aditof::Camera> &camera) = 0;
+    virtual void publishMsg(const ros::Publisher &pub) = 0;
+};
+
+inline AditofSensorMsg::~AditofSensorMsg() {}
```
dNechita: A much shorter way to to this is:
`virtual ~AditofSensorMsg() = default;`

bindings/ros/aditof_roscpp/src/message_factory.cpp
```diff
@@ -0,0 +1,13 @@
+#include "message_factory.h"
+
+AditofSensorMsg *
+MessageFactory::create(const std::shared_ptr<aditof::Camera> &camera,
+                       MessageType type) {
+    switch (type) {
+    case MessageType::sensor_msgs_PointCloud2:
+        return new PointCloud2Msg(camera);
+    default:
```
dNechita: Default might do more harm than good. I would remove it. This way if at some point in future one adds a new element to the enum but forgets to handle it in this switch, the compiler will tell us that. Otherwise the new element will fall to default and this will be probably detected at run-time, if lucky.

bindings/ros/aditof_roscpp/src/examples/rviz_pointcloud.cpp
```diff
@@ -0,0 +1,36 @@
...
+
+    ros::init(argc, argv, "aditof_rviz_node");
+
+    ros::NodeHandle nHandle;
+    ros::Publisher frame_pubisher =
+        nHandle.advertise<sensor_msgs::PointCloud2>("aditof_pcloud", 100);
+
+    applyNoiseReduction(camera, argc, argv);
+
+    AditofSensorMsg *pclMsg =
```
dNechita: when does this get deleted?

bindings/ros/aditof_roscpp/src/aditof_utils.cpp
andreeasandulescu: My mistake. I should have used the glog logging statements instead of the ones used in ros. Fixed it, thanks

bindings/ros/aditof_roscpp/src/pointcloud2_msg.cpp
andreeasandulescu: changed the function name to FrameDataToMsg

andreeasandulescu:
Addresses https://github.com/analogdevicesinc/aditof_sdk/issues/84

PR #216, opened 2020-02-20
Chicony support

dNechita:


sdk/src/camera_chicony.cpp
```diff
@@ -0,0 +1,252 @@
+#include "camera_chicony.h"
+
+#include <aditof/camera_chicony_specifics.h>
```
DanielGuramulta: This include already exists in `camera_chicony.h`

sdk/include/aditof/camera_chicony_specifics.h
```diff
@@ -0,0 +1,62 @@
...
+     * @return Status
+     */
+    Status setNoiseReductionThreshold(uint16_t threshold);
+
+    /**
+     * @brief Returns the last value that has been set for the threshold of the
+     * noise reduction.
+     * @return uint16_t
+     */
+    uint16_t noiseReductionThreshold() const;
```
DanielGuramulta: This method is not implemented in the `.cpp` file. Windows builds on Appveyor complain about this.

Issue #220, opened 2020-02-24
Bindings: Change coordinate system for ROS point cloud


andreeasandulescu:
The point cloud data should be specified relative to the world coordinate system

PR #222, opened 2020-02-24
bindings: Fix wrong coordinate system for ROS

andreeasandulescu:
Points contained in the Pointcloud message are defined relative to the world coordinate system


andreeasandulescu:
Fixes bug described in issue: https://github.com/analogdevicesinc/aditof_sdk/issues/220

bindings/ros/aditof_roscpp/src/pointcloud2_msg.cpp
```diff
@@ -31,59 +31,46 @@ void PointCloud2Msg::setDataMembers(const std::shared_ptr<Camera> &camera,
...
-            *iter_x = x_n;
-            *iter_y = y_n;
-            *iter_z = z_n;
-
-            *iter_rgb = frameData[i * msg.width + j];
+            int index = i * msg.width + j;
+            uint16_t depth = frameData[index];
+
+            // mark the points corresponding to the min and max range as invalid
+            if (depth == *minRange || depth == *maxRange) {
```
grigorovici-andreea: why do you exclude the edges?

andreeasandulescu: It is a feature that is not concerned with changing the coordinate system, so it should not have been included in this PR. Removed in commit https://github.com/analogdevicesinc/aditof_sdk/pull/222/commits/05c28a1a42f2e40c11ef847cec6d4574b89654c1

PR #223, opened 2020-02-27
scripts: Fix cmake prefix paths in Raspberry PI script

dNechita:


dNechita:
> The fix looks good, but do we still need this script? We have this one: https://github.com/analogdevicesinc/aditof_sdk/blob/master/scripts/raspberrypi3/setup.sh

Well, it's in the user guide: https://github.com/analogdevicesinc/aditof_sdk/blob/master/doc/raspberrypi3/user_guide.md

Issue #232, opened 2020-03-03
Demo: noise reduction, pressing backspace does not change the value of smallSignalThreshold


grigorovici-andreea:
If the value of smallSignalThreshold is changed by pressing the backspace, it does not compute. Other key needs to be pressed in order for this to work.

PR #233, opened 2020-03-04
Demo: solved noise reduction bug.

grigorovici-andreea:
The value of the small signal threshold it is changed if the backspace is pressed.

Solves: #232

### DIFF
--- a/examples/aditof-demo/aditofdemoview.cpp
+++ b/examples/aditof-demo/aditofdemoview.cpp
@@ -748,6 +748,9 @@ void AdiTofDemoView::render() {
                 }
             } else if (backspace) {
                 value = value.substr(0, value.size() - 1);
+                if (value.compare("") != 0) {
+                    smallSignalThreshold = stoi(value);
+                }
             }
         }
 


examples/aditof-demo/aditofdemoview.cpp
```diff
@@ -748,6 +748,7 @@ void AdiTofDemoView::render() {
                 }
             } else if (backspace) {
                 value = value.substr(0, value.size() - 1);
+                smallSignalThreshold = stoi(value);
```
dNechita: on the if branch, there is some validation to check if a value is not greater than (1 << 14) - 1.
What if you have a very large value and after pressing backspace the resulting value is still greater than that threshold? You're missing the validation here, on the else branch.

grigorovici-andreea: The value from the box cannot be greater than (1 << 14) - 1. This is because of the UI which is really rudimental and there is not copy=paste capability, and also the cursor cannot be moved in the box (so you can only delete from the last digit). Also, the maximum number that can be introduced in the box is (1 << 14) - 1.

I though solved another bug: if all the digits are deleted with backspace, the application should treat this case as if no noise reduction is requested.

PR #236, opened 2020-03-05
Demo: far mode does not support noise reduction.

grigorovici-andreea:
Because of hardware limitations, noise reduction does not work in far mode.
This is disabled from extended camera API.

### DIFF
--- a/sdk/include/aditof/status_definitions.h
+++ b/sdk/include/aditof/status_definitions.h
@@ -15,6 +15,7 @@ enum class Status {
     BUSY,             //!< Device or resource is busy
     UNREACHABLE,      //!< Device or resource is unreachable
     INVALID_ARGUMENT, //!< Invalid arguments provided
+    UNAVAILABLE,      //!< The requested action or resource is unavailable
     GENERIC_ERROR     //!< An error occured but there are no details available.
 };
 
--- a/sdk/src/camera_96tof1_specifics.cpp
+++ b/sdk/src/camera_96tof1_specifics.cpp
@@ -50,6 +50,13 @@ uint16_t Camera96Tof1Specifics::noiseReductionThreshold() const {
 }
 
 Status Camera96Tof1Specifics::setTresholdAndEnable(uint16_t treshold, bool en) {
+    aditof::CameraDetails cameraDetails;
+    m_camera->getDetails(cameraDetails);
+    if (cameraDetails.mode.compare("far") == 0) {
+        LOG(WARNING) << "Far mode does not support noise reduction!";
+        return Status::UNAVAILABLE;
+    }
+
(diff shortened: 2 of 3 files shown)


sdk/include/aditof/status_definitions.h
```diff
@@ -11,11 +11,12 @@ namespace aditof {
...
-    OK,               //!< Success
-    BUSY,             //!< Device or resource is busy
-    UNREACHABLE,      //!< Device or resource is unreachable
-    INVALID_ARGUMENT, //!< Invalid arguments provided
-    GENERIC_ERROR     //!< An error occured but there are no details available.
+    OK,                    //!< Success
+    BUSY,                  //!< Device or resource is busy
+    UNREACHABLE,           //!< Device or resource is unreachable
+    INVALID_ARGUMENT,      //!< Invalid arguments provided
+    NOISE_REDUCTION_ERROR, //!< Noise reduction is not possible on far mode
```
dNechita: This is too specific for an error! We can't create an error for each feature we do.
Use the logger to print a more detailed message.
Use something like NOT_AVAILABLE or something like that, which can be reused.

grigorovici-andreea: changed.

Thank you

sdk/src/camera_96tof1_specifics.cpp
```diff
@@ -50,6 +50,13 @@ uint16_t Camera96Tof1Specifics::noiseReductionThreshold() const {
 }
 
 Status Camera96Tof1Specifics::setTresholdAndEnable(uint16_t treshold, bool en) {
+    aditof::CameraDetails cameraDetails;
+    m_camera->getDetails(cameraDetails);
+    if (cameraDetails.mode.compare("far") == 0) {
+        LOG(ERROR) << "Far mode does not support noise reduction";
```
dNechita: This is treated too severely. A warning should be enough.

grigorovici-andreea: solved. thank you

sdk/include/aditof/status_definitions.h
```diff
@@ -15,6 +15,7 @@ enum class Status {
     BUSY,             //!< Device or resource is busy
     UNREACHABLE,      //!< Device or resource is unreachable
     INVALID_ARGUMENT, //!< Invalid arguments provided
+    NOT_AVAILABLE,    //!< The requested feature is not available
```
dNechita: A slightly shorter alternative is: UNAVAILABLE
The description could be more generic: "The requested action or resource is unavailable"

grigorovici-andreea: done.

Issue #237, opened 2020-03-06
MATLAB Adaptor build broken


tfcollins:
See here: https://gitlab.com/tfcollins/TimeofFlightToolbox/-/jobs/461639926#L4601

Looks like the aditof::CameraDetails struct changed.

DanielGuramulta:
Will have a look at the issue, and fix it as soon as possible.

Thanks for reporting it.

tfcollins:
@DanielGuramulta I'll add a daily build against master for MATLAB

PR #238, opened 2020-03-06
bindings: matlab: fix adaptor build

DanielGuramulta:
Build was failing due to a change in the structure of the camera details. Now the depth image should be converted to pixels according to the minimum and maximum depth value it can measure in the desired mode. 

Fixes issue: #237  


tfcollins:
Works on my CI now. Thanks!

PR #240, opened 2020-03-06
doc: Use 'halt' over 'poweroff' in dragonboard doc

dNechita:
Poweroff will turn off the board for a couple of seconds and reboot the board afterwards.
Halt doesn't reboot.

### DIFF
--- a/doc/dragonboard410c/user_guide.md
+++ b/doc/dragonboard410c/user_guide.md
@@ -19,7 +19,7 @@
 ![DB410C connections](https://github.com/analogdevicesinc/aditof_sdk/blob/master/doc/img/db410c_standalone.JPG)
 
 ### Power off sequence
-- under Linux open a terminal and type ***sudo poweroff***. This will safely power off the DragonBoard410c and ensure that the SD card is properly umounted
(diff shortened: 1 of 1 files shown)


doc/dragonboard410c/user_guide.md
```diff
@@ -19,16 +19,14 @@
 ![DB410C connections](https://github.com/analogdevicesinc/aditof_sdk/blob/master/doc/img/db410c_standalone.JPG)
 
 ### Power off sequence
-- under Linux open a terminal and type ***sudo poweroff***. This will safely power off the DragonBoard410c and ensure that the SD card is properly umounted
+- under Linux open a terminal and type ***sudo halt***. This will safely turn off the OS on DragonBoard410c and ensure that the SD card is properly umounted
```
grigorovici-andreea: Should be: **unmounted**

dNechita: The command that unmounts is called umount.
'Unmounts' is better, makes things more readable. This needs to be changed in multiple user guides, including the ones on wiki.analog but it's not the purpose of this PR.

PR #241, opened 2020-03-06
Provide CI for ROS package

andreeasandulescu:
Added two more jobs in travis.yml file, for the latest ROS distributions.
Building the ROS package is enabled by setting a cmake parameter, similarly to the other bindings.

### DIFF
--- a/bindings/ros/CMakeLists.txt
+++ b/bindings/ros/CMakeLists.txt
@@ -1,0 +1,24 @@
+cmake_minimum_required(VERSION 2.8.3)
+
+if(UNIX)
+
+  # set variables
+  set(CATKIN_WS "${CMAKE_BINARY_DIR}/catkin_ws")
+  set(ROS_SRC_DIR "${CMAKE_CURRENT_SOURCE_DIR}")
+
+  # create and run ROS setup script
+  configure_file(${CMAKE_CURRENT_SOURCE_DIR}/ros_setup.bash
+                 ${CMAKE_BINARY_DIR}/tmp/ros_setup.bash)
+  file(
+    COPY ${CMAKE_BINARY_DIR}/tmp/ros_setup.bash
+    DESTINATION ${CMAKE_BINARY_DIR}
+    FILE_PERMISSIONS OWNER_READ OWNER_WRITE OWNER_EXECUTE)
+
+  add_custom_target(aditof_ros_package)
+  string(CONCAT OLD_CMAKE_INSTALL_PREFIX "\"${CMAKE_INSTALL_PREFIX}\"")
+  string(CONCAT OLD_CMAKE_PREFIX_PATH "\"${CMAKE_PREFIX_PATH}\"")
+  add_custom_command(
+    TARGET aditof_ros_package
+    COMMAND ${CMAKE_BINARY_DIR}/ros_setup.bash "${OLD_CMAKE_INSTALL_PREFIX}"
+            "${OLD_CMAKE_PREFIX_PATH}" ${VERSION} DEPENDS aditof)
+endif()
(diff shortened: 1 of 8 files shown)


bindings/ros/aditof_roscpp/CMakeLists.txt
```diff
@@ -1,31 +1,34 @@
...
+# Find catkin macros and libraries
+find_package(catkin REQUIRED COMPONENTS roscpp sensor_msgs)
 
-catkin_package(
-  INCLUDE_DIRS include
-  CATKIN_DEPENDS roscpp sensor_msgs
-)
+# if we create the catkin_ws directory inside the build directory, find_package
+# is not needed
+if(CMAKE_SOURCE_DIR MATCHES ".*aditof_sdk/build.*")
```
dNechita: There is no guarantee that build folder is going to be named 'build'. The cmake should not assume that.

bindings/ros/CMakeLists.txt
```diff
@@ -0,0 +1,27 @@
+cmake_minimum_required(VERSION 2.8.3)
+
+if(UNIX)
+
+	#set variables
+	if(NOT DEFINED CATKIN_WS)
+		message("-- Creating catkin workspace in build directory")
```
dNechita: Are these messages useful for user? Or they acted as debug messages while writing and testing these cmake instructions?

andreeasandulescu: Generally, ros packages are generated in a catkin workspace, so we decided that there are two use cases for the ros bindings:
1.A catkin workspace directory is created in the build directory of aditof_sdk 
2.The user has already created a catkin workspace
For the second case, the user must provide the path of the workspace using the CATKIN_WS variable. The messages were meant for users, to make it easier for them to check what is the path to the catkin workspace used, maybe warn them that they forgot to set the parameter. What is your opinion? Are the messages useful or not?

dNechita: In case 1. things are straightforward, as one can see that everything (including the ros packages) is created in the build folder.
In case 2. the user provides the path. Printing the path that user provided isn't of much help. If user forgets to provide path, the cmake should fail as it can't continue anymore, right?

andreeasandulescu: You are right, but how to tell if the user didn't provide a path because he wants to follow the first case or because he wants to follow the second case and forgot? First option, the cmake shouldn't fail, but the second should. Would it be ok for the user to set the CATKIN_WS variable to "default" or something similar for the first case and an unset variable would result in an error message and cmake failing?

dNechita: What if user forgets to unset?
You should detect automatically if you're in case 1 or case 2.

bindings/ros/aditof_roscpp/CMakeLists.txt
andreeasandulescu: Fixed in commit https://github.com/analogdevicesinc/aditof_sdk/pull/241/commits/12add7a2dcb99d8987e9e3b1573f21e1e1957c76

bindings/ros/CMakeLists.txt
andreeasandulescu: As discussed, the second case is no longer supported. Fixed in commit https://github.com/analogdevicesinc/aditof_sdk/pull/241/commits/12add7a2dcb99d8987e9e3b1573f21e1e1957c76

andreeasandulescu:
Solves issue: https://github.com/analogdevicesinc/aditof_sdk/issues/234
As discussed, there is only one new job instead of two, for the latest ROS distribution
The CMakeLists.txt file in the ros directory saves the values of the variables CMAKE_INSTALL_PREFIX and CMAKE_PREFIX_PATH because after the commands "source /opt/ros/$ROS_DISTRO/setup.sh" and catkin_make, their values will change.

Issue #165, opened 2019-11-29
Start network server on target by pressing S4


acozma:
Start network server on target by pressing S4 on the ToF camera board.
The red DS5 led should light up when the server starts.
USB and ETH are mutually exclusive.

DanielGuramulta:
Should investigate if we could start the network server by default.

dNechita:
Check auto-detection on network.

dNechita:
Start server by default on all targets.

PR #247, opened 2020-03-11
utils/raspberrypi: Add service for starting tof server during boot

dNechita:
Handles #165 (the part for Raspberry PI)


utils/raspberrypi/readme.md
```diff
@@ -0,0 +1,26 @@
+# Utilities for Raspberry Pi
+
+This directory contains different utilities may be needed on the Raspberry Pi.
```
andreeasandulescu: Shouldn't it be "utilities that/which may be needed"?

dNechita: yes, thanks

PR #250, opened 2020-03-11
bindings: ros: Add support for camera connected over USB

andreeasandulescu:
The default connection type is now USB.

### DIFF
--- a/bindings/ros/aditof_roscpp/src/aditof_utils.cpp
+++ b/bindings/ros/aditof_roscpp/src/aditof_utils.cpp
@@ -11,13 +11,14 @@ std::shared_ptr<Camera> initCameraEthernet(int argc, char **argv) {
     google::InitGoogleLogging(argv[0]);
     FLAGS_alsologtostderr = 1;
     Status status = Status::OK;
+    std::string ip;
 
     if (argc < 2) {
-        LOG(ERROR) << "No ip provided!";
-        return nullptr;
+        LOG(INFO) << "No ip provided, attempting to connect to the camera "
+                     "through USB";
+    } else {
+        ip = argv[1];
     }
-
-    std::string ip = argv[1];
 
     System system;
     status = system.initialize();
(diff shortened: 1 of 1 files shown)


bindings/ros/aditof_roscpp/src/aditof_utils.cpp
```diff
@@ -11,13 +11,14 @@ std::shared_ptr<Camera> initCameraEthernet(int argc, char **argv) {
     google::InitGoogleLogging(argv[0]);
     FLAGS_alsologtostderr = 1;
     Status status = Status::OK;
-
+    std::string ip;
     if (argc < 2) {
-        LOG(ERROR) << "No ip provided!";
-        return nullptr;
+        LOG(WARNING) << "No ip provided!";
```
dNechita: I think INFO is better suited here. Client gets informed that in absence of an IP, an attempt to connect to the camera through USB will be made. Which means the message should be updated to explain these things.

bindings/ros/aditof_roscpp/src/aditof_utils.cpp
```diff
@@ -27,7 +28,14 @@ std::shared_ptr<Camera> initCameraEthernet(int argc, char **argv) {
     }
 
     std::vector<std::shared_ptr<Camera>> cameras;
-    system.getCameraListAtIp(cameras, ip);
+    if (argc < 2) {
```
dNechita: Why not check if ip is not empty?

bindings/ros/aditof_roscpp/src/aditof_utils.cpp
andreeasandulescu: You're right. Fixed in commit https://github.com/analogdevicesinc/aditof_sdk/pull/250/commits/9f03a7a6556856aa50f986fbbbf5b0ceffd0dab8

bindings/ros/aditof_roscpp/src/aditof_utils.cpp
andreeasandulescu: Fixed in commit https://github.com/analogdevicesinc/aditof_sdk/pull/250/commits/9f03a7a6556856aa50f986fbbbf5b0ceffd0dab8

PR #266, opened 2020-03-18
scripts/windows Bugs fixes.

grigorovici-andreea:
Fixed bug: if you selected your own folders script was failing.
Fixed bug: help mentioned parameters that were not used.

### DIFF
--- a/scripts/windows/readme.md
+++ b/scripts/windows/readme.md
@@ -10,14 +10,6 @@ Please use option -h or --help for further reference on how to change the defaul
 
 * CMake (Windows installer can be downloaded from: https://cmake.org/download/)
 
-* OpenCV
-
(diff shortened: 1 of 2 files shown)


dNechita:
I tried this script and fails when it runs cmake for the aditof_sdk. It cannot find OpenCV package even though I have provided the opencv 3.4.1 root absolute path as requested by the script.

grigorovici-andreea:
- add in the script: set environment variables for OpenCV;
- add a tutorial for how to open the solution with visual studio.

grigorovici-andreea:
@dNechita 
> I tried this script and fails when it runs cmake for the aditof_sdk. It cannot find OpenCV package even though I have provided the opencv 3.4.1 root absolute path as requested by the script.

Should be fixed with now.

PR #273, opened 2020-03-24
bindings: ros: Add documentation

andreeasandulescu:
Initial version of README.md for ros.

### DIFF
--- a/bindings/ros/README.md
+++ b/bindings/ros/README.md
@@ -1,0 +1,40 @@
+# ROS Wrapper for the ADI ToF library
+
+## Overview
+This ROS package facilitates depth and IR data acquisition and processing for the Analog Devices depth cameras.
+
+## Installation
+
+- **Install the recommended [ROS distribution](http://wiki.ros.org/Distributions) for your operating system**
+  - [ROS Install page](http://wiki.ros.org/ROS/Installation)
+- **Make sure you have these ROS packages installed before building and running the examples**
+  - [rviz](http://wiki.ros.org/rviz)\
+    Run the command below for each package, replacing ROSDISTRO with the name of the ROS distribution you are using and PACKAGE with the name of the needed package.
+  ```console
+  sudo apt install ros-ROSDISTRO-PACKAGE
+  ```
+- **Install the ADI ToF SDK library**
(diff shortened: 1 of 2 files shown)


bindings/ros/README.md
```diff
@@ -0,0 +1,40 @@
+# ROS Wrapper for the ADI ToF library
+
+## Overview
+This ROS package facilitates depth and IR data acquisition and processing for the Analog Devices depth cameras.
+
+## Installation
+
+- **Install the recommended [ROS distribution](http://wiki.ros.org/Distributions) for your operating system**
+  - [ROS Install page](http://wiki.ros.org/ROS/Installation)
+- **[Optional] Install additional ROS packages needed to build and run the examples**
```
andreeasandulescu: I'm not sure if the "[Optional]" tag is a good choice, please let me know if you have a better idea.
Its purpose is to let the users know that they need to check if these packages are installed , or install them otherwise .I also chose the word optional because in this case, the rviz package is included in the Debian package ros-melodic-desktop-full needed to install ros, but it is not included in the ros base(Bare Bones) package ros-melodic-ros-base.

bindings/ros/README.md
```diff
@@ -0,0 +1,40 @@
+# ROS Wrapper for the ADI ToF library
+
+## Overview
+This ROS package facilitates depth and IR data acquisition and processing for the Analog Devices depth cameras.
+
+## Installation
+
+- **Install the recommended [ROS distribution](http://wiki.ros.org/Distributions) for your operating system**
```
grigorovici-andreea: do these steps work for windows too? Or is there something different that the user needs to do if he/she is on Windows?

bindings/ros/README.md
dNechita: Why not just say: "Install rviz if not already installed"?

dNechita:
@andreeasandulescu please address the questions made for this PR.

bindings/ros/README.md
andreeasandulescu: This ROS package does not offer support for Windows yet.

bindings/ros/README.md
andreeasandulescu: Because the list of packages may get longer.I think I would better change it to "Make sure you have these packages installed before you run the examples".

PR #312, opened 2020-04-23
Add image msg

andreeasandulescu:
- **bindings: ros: Code refactor for frame-based message:** Creating a message no longer implies requesting a new frame.
- **bindings: ros: Use IR data to color point cloud:** Set intensity channel values based on the IR data.
- **bindings: ros: Add image messages:** Add two new types of messages, based on sensor_msgs/Image.

### DIFF
--- a/bindings/ros/aditof_roscpp/src/examples/rviz_pointcloud.cpp
+++ b/bindings/ros/aditof_roscpp/src/examples/rviz_pointcloud.cpp
@@ -37,11 +37,14 @@ using namespace aditof;
 
 int main(int argc, char **argv) {
 
-    std::shared_ptr<Camera> camera = initCameraEthernet(argc, argv);
+    std::shared_ptr<Camera> camera = initCamera(argc, argv);
     if (camera.get() == nullptr) {
         ROS_ERROR("initCamera call failed");
         return 0;
     }
+
+    setFrameType(camera, "depth_ir");
+    setMode(camera, "medium");
 
     ros::init(argc, argv, "aditof_rviz_node");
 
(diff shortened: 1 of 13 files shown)


bindings/ros/aditof_roscpp/src/examples/rviz_pointcloud.cpp
```diff
@@ -37,12 +37,15 @@ using namespace aditof;
 
 int main(int argc, char **argv) {
 
-    std::shared_ptr<Camera> camera = initCameraEthernet(argc, argv);
+    std::shared_ptr<Camera> camera = initCamera(argc, argv);
     if (camera.get() == nullptr) {
         ROS_ERROR("initCamera call failed");
         return 0;
```
grigorovici-andreea, line 43: Usually if return is 0, that means is with success, but you call an error before. in my opinion it is not consistent.

dNechita:
The title of the PR says: "Add image msg" but then there are other commits which do more things: refactoring and coloring the pointcloud. These should be in separate PR.

bindings/ros/aditof_roscpp/src/examples/rviz_pointcloud.cpp
```diff
@@ -37,12 +37,15 @@ using namespace aditof;
 
 int main(int argc, char **argv) {
 
-    std::shared_ptr<Camera> camera = initCameraEthernet(argc, argv);
+    std::shared_ptr<Camera> camera = initCamera(argc, argv);
     if (camera.get() == nullptr) {
```
DanielGuramulta, line 41: The if condition can be written in a simpler way
```cpp
if (!camera) {
    ...
```

Issue #306, opened 2020-04-14
MaskR-CNN cpp


grigorovici-andreea:
Try to implement maskR-CNN OpenCV example in CPP.

PR #313, opened 2020-04-23
opencv/maskrcnn: Added maskR-CNN example in Cpp

grigorovici-andreea:
Added maskR-CNN example in Cpp using OpenCV
Added documentation and output example.

Solves: #306 


bindings/opencv/maskrcnn/CMakeLists.txt
```diff
@@ -0,0 +1,18 @@
+cmake_minimum_required(VERSION 2.8)
+
+project(aditof-opencv-maskrcnn)
+
+add_executable(${PROJECT_NAME} main.cpp)
+
+file(COPY ${CMAKE_CURRENT_SOURCE_DIR}/frozen_inference_graph.pb
```
DanielGuramulta, line 7: Can't we just give the user some documentation on how to get this files? They seem a bit big to be included in the repo and could prolong the duration of `cloning` the repo even when this example is not needed. (frozen_inference_graph.pb is 64MB)

bindings/opencv/CMakeLists.txt
```diff
@@ -29,7 +29,12 @@ if(OPENCV_FOUND OR
     else()
         message(STATUS "The dnn example will be built (OpenCV version >= 3.4.1)")
         add_subdirectory(dnn)
-
+        if(OpenCV_VERSION VERSION_LESS "3.4.3")
+            message(WARNING "The maskR-CNN example is only available for OpenCV version >= 3.4.3! This example won't be built!")
+        else()
+            message(STATUS "The maskR-CNN example will be built (OpenCV version >= 3.4.3)")
```
dNechita, line 35: I don't think it's necessary to print this.

bindings/opencv/maskrcnn/CMakeLists.txt
dNechita: If the files are binary and/or have large sizes, then we shouldn't include them in the repository.

bindings/opencv/CMakeLists.txt
grigorovici-andreea: For DNN example both of these messages are printed also. I think this should be consistent. Either I delete from both, either I let it this way. What do you think?

dNechita: delete both

bindings/opencv/maskrcnn/CMakeLists.txt
grigorovici-andreea: solved.

bindings/opencv/CMakeLists.txt
grigorovici-andreea: solved.

PR #315, opened 2020-04-23
bindings: ros: Code refactor for frame-based message

andreeasandulescu:
An AditofSensorMsg object is now created based on a Frame object.
In the previous version, creating a new message implied requesting
a new frame, which is incorrect.
This change is needed because the camera node will create different
types of messages based on the same frame.

### DIFF
--- a/bindings/ros/aditof_roscpp/src/examples/rviz_pointcloud.cpp
+++ b/bindings/ros/aditof_roscpp/src/examples/rviz_pointcloud.cpp
@@ -37,11 +37,14 @@ using namespace aditof;
 
 int main(int argc, char **argv) {
 
-    std::shared_ptr<Camera> camera = initCameraEthernet(argc, argv);
-    if (camera.get() == nullptr) {
+    std::shared_ptr<Camera> camera = initCamera(argc, argv);
+    if (!camera) {
         ROS_ERROR("initCamera call failed");
-        return 0;
+        return -1;
     }
+
+    setFrameType(camera, "depth_ir");
+    setMode(camera, "medium");
 
     ros::init(argc, argv, "aditof_rviz_node");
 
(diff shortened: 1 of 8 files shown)


bindings/ros/aditof_roscpp/src/examples/rviz_pointcloud.cpp
```diff
@@ -51,19 +54,29 @@ int main(int argc, char **argv) {
...
     while (ros::ok()) {
-        dynamic_cast<PointCloud2Msg *>(pclMsg)->FrameDataToMsg(camera);
+        getNewFrame(camera, &frame);
+        PointCloud2Msg *pclMsg = dynamic_cast<PointCloud2Msg *>(pclMsg);
+
+        if (!pclMsg) {
+            ROS_ERROR("downcast from AditofSensorMsg to PointCloud2Msg failed");
+            return -1;
+        }
+        dynamic_cast<PointCloud2Msg *>(pclMsg)->FrameDataToMsg(camera, &frame);
```
DanielGuramulta, line 76: Is this dynamic_cast necessary? You are already down-casting on line 70
```cpp
PointCloud2Msg *pclMsg = dynamic_cast<PointCloud2Msg *>(pclMsg);
```
so `pclMsg` is already of type `PointCloud2Msg`.
Also shouldn't it be `dynamic_cast<PointCloud2Msg *>(msg)` instead of `dynamic_cast<PointCloud2Msg *>(pclMsg)`. Maybe I'm missing something

andreeasandulescu: You're right. I made the changes requested in previous PR comments to replace line 76 and then forgot to delete that line. Fixed now, thanks.

PR #318, opened 2020-04-24
Added the IR gamma conversion function

acozma:
This function is used to control the contrast of the IR image

### DIFF
--- a/sdk/src/camera_96tof1_specifics.cpp
+++ b/sdk/src/camera_96tof1_specifics.cpp
@@ -31,6 +31,7 @@
  */
 #include <aditof/camera_96tof1_specifics.h>
 #include <glog/logging.h>
+#include <math.h>
 
 #include "camera_96tof1.h"
 
--- a/sdk/src/camera_96tof1_specifics.cpp
+++ b/sdk/src/camera_96tof1_specifics.cpp
@@ -107,16 +107,16 @@ Status Camera96Tof1Specifics::setIrGammaConversion(float gamma) {
     uint16_t y_val[9];
 
     for (int i = 0; i < 9; i++) {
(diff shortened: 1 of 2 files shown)


sdk/src/camera_96tof1_specifics.cpp
```diff
@@ -99,3 +100,26 @@ Status Camera96Tof1Specifics::setTresholdAndEnable(uint16_t treshold, bool en) {
...
+    for (int i = 0; i < 9; i++) {
+		y_val[i] = (uint16_t)(pow(x_val[i] / 4096.0f, gamma)* 1024.0f);
+    }
+
+    uint16_t afeRegsAddr[] = {0x4001, 0x7c22, 0xc372, 0xc373, 0xc374, 0xc375, 0xc376, 0xc377,
+                              0xc378, 0xc379, 0xc37a, 0xc37b, 0xc37c, 0xc37d, 0x4001, 0x7c22};
+    uint16_t afeRegsVal[]  = {0x0006, 0x0004, 0x7888, 0xa997, 0x000a, y_val[0], y_val[1], y_val[2],
+                              y_val[3],y_val[4], y_val[5], y_val[6], y_val[7], y_val[8], 0x0007, 0x0004};
+
+    status = m_camera->m_device->writeAfeRegisters(afeRegsAddr, afeRegsVal, 8);
```
dNechita, line 118: Why not write all 16 registers at once?

dNechita:
These changes need formatting.

sdk/src/camera_96tof1_specifics.cpp
acozma: For this AFE setting there seems to be an issue on USB with writing all the registers at once, maybe because of some delays in the AFE

dNechita: I think a comment that says this would be very useful.

PR #321, opened 2020-04-28
bindings: ros: Use IR data to color point cloud

andreeasandulescu:
Use intensity channel instead of rgb channel and set
the values accordingly (IR data instead of depth data).

### DIFF
--- a/bindings/ros/aditof_roscpp/src/aditof_utils.cpp
+++ b/bindings/ros/aditof_roscpp/src/aditof_utils.cpp
@@ -188,3 +188,20 @@ int getRangeMin(const std::shared_ptr<Camera> &camera) {
     camera->getDetails(cameraDetails);
     return cameraDetails.minDepth;
 }
+
+void irTo16bitGrayscale(uint16_t *frameData, int width, int height) {
+    std::vector<uint16_t> data(frameData, frameData + width * height);
+
+    auto min_val = std::min_element(data.begin(), data.end());
+    auto max_val = std::max_element(data.begin(), data.end());
+    uint16_t delta = *max_val - *min_val;
+    int minColorValue = 0;
+
+    for (int i = 0; i < width * height; i++) {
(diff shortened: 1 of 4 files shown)


bindings/ros/aditof_roscpp/src/aditof_utils.cpp
```diff
@@ -188,3 +188,19 @@ int getRangeMin(const std::shared_ptr<Camera> &camera) {
...
+
+    auto min_val = std::min_element(data.begin(), data.end());
+    auto max_val = std::max_element(data.begin(), data.end());
+    uint16_t delta = *max_val - *min_val;
+
+    for (int i = 0; i < width * height; i++) {
+        float norm_val = static_cast<float>(data[i] - *min_val) / delta;
+        float grayscale_val =
+            norm_val * std::numeric_limits<unsigned short int>::max() +
+            (1.0f - norm_val) * 0;
```
grigorovici-andreea, line 203: This seems weird to me: (1.0f - norm_val) * 0 . Why would you do some calculations to multiply afterward with 0. this is going to be 0 anyway.

andreeasandulescu: This is used to map the color between a max and a min value. I kept the formula as is because it seems clearer to me that the color is mapped in the interval [0,  maxValue].In the future, the min value could b changed from 0 to some other value, in which case the part you pointed out shouldn't be deleted.

grigorovici-andreea: Then, I think you should've created a variable before, with a good name, and initialize that variable with 0, and in the future you can change that variable. This way is a lot easier for other people to understand it.

andreeasandulescu: @dNechita What do you think?

andreeasandulescu: Changes applied, thanks!

PR #325, opened 2020-04-28
bindings: ros: Add image and camera_info messages

andreeasandulescu:
Added  new classes that encapsulate the CameraInfo and
Image messages defined in the ros package sensor_msgs.
There are two types for the Image message, an 8-bit
RGB depth image and a 16-bit grayscale IR one.


tamaslevente:
Camera_info currently contains a small bug in constructing the K and P matrices, Check the reference at:
http://docs.ros.org/indigo/api/sensor_msgs/html/msg/CameraInfo.html

tamaslevente:
Depth message representation seems to be more appropriate with TYPE_16UC1 encoding, as this is the preferred way of storing depth images:
http://docs.ros.org/jade/api/depth_image_proc/html/point__cloud__xyz_8cpp_source.html

Camera_info should be in synch with depth/ir (i.e. same time stamp), else the image processing nodes (e.g. image_proc) cannot handle them.

PR #326, opened 2020-04-28
sdk: target: local_device.cpp: check if buf.bytesused different from 0

DanielGuramulta:
Seems like buf.bytesused is 0 on the dragonboard target when in fact it should be (width * height * 3 / 2) as data for the frame is packed.

### DIFF
--- a/sdk/src/target/local_device.cpp
+++ b/sdk/src/target/local_device.cpp
@@ -538,6 +538,18 @@ aditof::Status LocalDevice::getFrame(uint16_t *buffer) {
         return status;
     }
 
+    auto isBufferPacked = [](const struct v4l2_buffer &buf, unsigned int width,
+                             unsigned int height) {
+        unsigned int bytesused = 0;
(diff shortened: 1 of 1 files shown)


btogorean:
According to https://www.kernel.org/doc/html/v4.12/media/uapi/v4l/buffer.html#c.v4l2_buffer if format type is `multiplanar` this field is ignored. I think you should try to get the information from `buf.m.planes[0].bytesused`

DanielGuramulta:
> According to https://www.kernel.org/doc/html/v4.12/media/uapi/v4l/buffer.html#c.v4l2_buffer if format type is `multiplanar` this field is ignored. I think you should try to get the information from `buf.m.planes[0].bytesused`

Thanks! Will have a look at the code and change it to take into account the fact that some buffers might be multiplanar

DanielGuramulta:
@btogorean Implemented your approach for multiplanar buffers. It fixes the problem on the dragonboard platform!. Can you test this branch on the Jetson just to make sure it doesn't break anything?

PR #354, opened 2020-05-21
sdk: jetson: imshow-jetson: Fix seg fault with chicony cam

btogorean:
On imshow-jetson example there is generated a segmentation fault
when chicony camera is used due to returned null pointer.
This commit fix that.

Signed-off-by: Bogdan Togorean <bogdan.togorean@analog.com>


btogorean:
Changes in v2:
- commit author is now correct

PR #357, opened 2020-05-25
Demo: Ui fixes - gamma correction

grigorovici-andreea:
Added length limit for gamma correction value input.
Made sure the user cannot input for gamma correction ".", as a start value, as this generates a fatal error.
Made sure only one "." can be inputed for the gamma correction value.

### DIFF
--- a/examples/aditof-demo/aditofdemoview.cpp
+++ b/examples/aditof-demo/aditofdemoview.cpp
@@ -844,8 +844,15 @@ void AdiTofDemoView::render() {
 
         if (valueFieldSelectedG) {
             if ((key >= 48 && key <= 57) || (key == 46)) {
-                valueG += pressedValidKey;
-                IRGamma = stof(valueG);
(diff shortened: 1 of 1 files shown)


examples/aditof-demo/aditofdemoview.cpp
```diff
@@ -844,8 +844,13 @@ void AdiTofDemoView::render() {
 
         if (valueFieldSelectedG) {
             if ((key >= 48 && key <= 57) || (key == 46)) {
-                valueG += pressedValidKey;
-                IRGamma = stof(valueG);
+                if ((valueG.size() < 11) &&
```
dNechita, line 847: Wouldn't it make sense to try to convert the newly formed string to float and if that fails, invalidate the key? And avoid all this complicated logic?

grigorovici-andreea: modified. thanks

Issue #358, opened 2020-05-26
Fix matrices in ROS CameraInfo message


andreeasandulescu:

andreeasandulescu:
Camera_info currently contains a small bug in constructing the K and P matrices, Check the reference at:
http://docs.ros.org/indigo/api/sensor_msgs/html/msg/CameraInfo.html

_Originally posted by @tamaslevente in https://github.com/analogdevicesinc/aditof_sdk/pull/325#issuecomment-633283659_

PR #359, opened 2020-05-26
bindings: ros: Fix camera_info msg matrix

andreeasandulescu:
The K matrix contained in the camera_info message was missing a row.


andreeasandulescu:
Fixes #358

Issue #362, opened 2020-05-27
Add make option to build SDK without Ethernet support


acozma:
This will help reduce the dependencies to websockets and protobuf

bitslip6:
Would love this feature :)

DanielGuramulta:
Enhancement implemented with pr: #375

PR #375, opened 2020-06-11
Disable network interface

DanielGuramulta:
Feature mentioned in issue: #362


sdk/include/aditof/device_construction_data.h
```diff
@@ -44,9 +44,11 @@ namespace aditof {
  * @brief Provides the types of sensor assosiated with the device
  */
 enum class DeviceType {
-    LOCAL,    //!< on the target
-    USB,      //!< connects to target via USB
+    LOCAL, //!< on the target
+    USB,   //!< connects to target via USB
+#ifdef HAS_NETWORK
```
dNechita, line 49: API shouldn't change based on some build option.

sdk/src/device_enumerator_ethernet.cpp
```diff
@@ -30,15 +30,18 @@
  * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
```
dNechita, line 30: device_enumerator_ethernet.h/cpp could be excluded by cmake when building without Network support.

sdk/src/device_factory.cpp
```diff
@@ -45,9 +49,11 @@ DeviceFactory::buildDevice(const aditof::DeviceConstructionData &data) {
     case DeviceType::USB: {
         return std::unique_ptr<DeviceInterface>(new UsbDevice(data));
     }
+#ifdef HAS_NETWORK
     case DeviceType::ETHERNET: {
         return std::unique_ptr<DeviceInterface>(new EthernetDevice(data));
```
dNechita, line 54: Should report at runtime that network support is disabled.

DanielGuramulta:
@dNechita I have addressed all the issues you mentioned and fixed them:
1) The api is no longer changed(`findDevices` method and `DeviceType` enum)
2) Moved the error message that should be seen at runtime in the systems implementation. How this works: when the user calls `getCameraAtIp(...)` and the sdk was built with `WITH_NETWORK=off` an error message will be logged telling the user that the network interface is disabled and to rebuild with the correct options enabled.

Thank you!

Issue #374, opened 2020-06-11
Create kernel patches for the DragonBoard410c kernel to distribute instead of the full kernel repo


acozma:
Delete the aditof_linux repo at the end and update the DragonBoard410c Build instructions documentations to point to the kernel patches.

PR #377, opened 2020-06-16
Dgb410c image creation

dNechita:
Implements #374

### DIFF
--- a/sdcard-images-utils/README.md
+++ b/sdcard-images-utils/README.md
@@ -1,0 +1,6 @@
+# Utilities to create SD card images for different embedded platforms
+
+## Directory Structure
+| Directory | Description |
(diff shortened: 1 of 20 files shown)


DanielGuramulta:
wouldn't make sense to have all this scripts and patches in the `utils/dragonboard` folder instead of creating a new one?

dNechita:
> wouldn't make sense to have all this scripts and patches in the `utils/dragonboard` folder instead of creating a new one?

I would keep everything that is related to the creation of  SD card images separate from the rest. This way it is easier to manage all the scripts and patches instead of having them all mixed. In utils/dragonboard we have udev rules and services which we want to be copied in the system when installing the sdk. The same does not apply to the image creating scripts.

PR #383, opened 2020-06-24
bindings/python fixed errors.

grigorovici-andreea:
Errors resulted from replacing cameraSpecifics with controls in sdk.

### DIFF
--- a/bindings/python/examples/showPointCloud/showPointCloud.py
+++ b/bindings/python/examples/showPointCloud/showPointCloud.py
@@ -78,9 +78,6 @@ if __name__ == "__main__":
     if not status:
         print("cameras[0].setFrameType() failed with status:", status)
 
-    specifics = cameras[0].getCamera96Tof1Specifics()
-    specifics.setCameraRevision(tof.Revision.RevC)
-
     status = cameras[0].setMode(modes[ModesEnum.MODE_NEAR.value])
     if not status:
         print("cameras[0].setMode() failed with status: ", status)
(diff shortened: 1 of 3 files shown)


bindings/python/examples/showPointCloud/showPointCloud.py
```diff
@@ -78,9 +78,6 @@ def transform_image(np_image):
     if not status:
         print("cameras[0].setFrameType() failed with status:", status)
 
-    specifics = cameras[0].getCamera96Tof1Specifics()
-    specifics.setCameraRevision(tof.Revision.RevC)
```
DanielGuramulta, line 82: Is there no longer an option to specify the revision or is `tof.Revision.RevC` the value set by default for the control?

grigorovici-andreea: there is but I wanted to be consistent with the other examples so I decided to take it out as it is not the purpose of the example. Do you think it is better to include it? 

For the other examples, only the demo one has this option to set the revision. So if I will include this I would rather do it in all the examples including the cpp ones (opencv, open3d, maybe even matlab).

grigorovici-andreea: Also, the default revision is C. So only is the user wants to change it to B, but I still do not think it should fit here.

dNechita: The default is RevC (https://github.com/analogdevicesinc/aditof_sdk/blob/master/sdk/src/cameras/ad-96tof1-ebz/camera_96tof1.cpp#L67).
I think there's no point in dragging these lines of code (that set the revision) into all examples.

dNechita:
Looks good, thanks!

PR #385, opened 2020-07-05
Fix build for WITH_NETWORK=OFF

acozma:
Fix build error when WITH_NETWORK=OFF - with the existing implementation only the files for network_device were being removed from the build list.

### DIFF
--- a/sdk/CMakeLists.txt
+++ b/sdk/CMakeLists.txt
@@ -54,8 +54,6 @@ endif()
 file(GLOB PLATFORM_SOURCES
     ${CMAKE_CURRENT_SOURCE_DIR}/src/${OS_SPECIFIC_DIR}/*.cpp
     ${CMAKE_CURRENT_SOURCE_DIR}/src/${OS_SPECIFIC_DIR}/*.c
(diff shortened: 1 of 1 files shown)


sdk/CMakeLists.txt
```diff
@@ -68,24 +66,30 @@ file(GLOB PLATFORM_HEADERS
...
+
 if ( ${OS_SPECIFIC_DIR} STREQUAL ${TARGET_STRING} )
     list(REMOVE_ITEM SOURCES ${CMAKE_CURRENT_SOURCE_DIR}/src/local_device.cpp)
 endif()
 
-if ( NOT WITH_NETWORK )
-    list (REMOVE_ITEM SOURCES ${CMAKE_CURRENT_SOURCE_DIR}/src/connections/network/network.cpp)
-    list (REMOVE_ITEM SOURCES ${CMAKE_CURRENT_SOURCE_DIR}/src/connections/network/network.h)
+#if ( NOT WITH_NETWORK )
     list (REMOVE_ITEM SOURCES ${CMAKE_CURRENT_SOURCE_DIR}/src/ethernet_device.cpp)
```
dNechita, line 82: This always removes ethernet_device.cpp/h. All builds are failing because of this.

PR #398, opened 2020-09-02
Api2.0

dNechita:
This splits the DeviceIntreface and creates the EepromInterface where everything related to EEPROM is moved from DeviceInterface to EepromInterface.


apps/server/server.cpp
```diff
@@ -511,6 +505,155 @@ void invoke_sdk_api(payload::ClientRequest buff_recv) {
...
+        buff_send.set_status(static_cast<::payload::Status>(status));
+        break;
+    }
+
+    case EEPROM_READ: {
+#ifdef DEBUG
+        cout << "EepromRead function\n";
+#endif
+        // Check if client has sent exactly one name to identify the EEPROM
+        if (buff_recv.device_data().eeproms_size() != 1) {
```
DanielGuramulta, line 554: Looks like this eeprom check line 554-572 is used in many places: `EEPROM_WRITE, READ and CLOSE case`, can this code be factored out in a function?

apps/server/server.cpp
```diff
@@ -511,6 +505,155 @@ void invoke_sdk_api(payload::ClientRequest buff_recv) {
...
+        }
+        std::string eName = buff_recv.device_data().eeproms(0).driver_name();
+
+        // We lookup the EEPROM instance by name
+        auto eeprom = findEepromByName(eName);
+        if (!eeprom) {
+            buff_send.set_message("Failed to idenify EEPROM with name: " +
+                                  eName);
+            buff_send.set_status(static_cast<::payload::Status>(
+                aditof::Status::INVALID_ARGUMENT));
```
DanielGuramulta, line 571: missing a `break;`? if there is no eeprom this line `577: aditof::Status status = eeprom->read(address, buffer, length);` will dereference a nullptr?

apps/server/server.cpp
```diff
@@ -511,6 +505,155 @@ void invoke_sdk_api(payload::ClientRequest buff_recv) {
...
+        }
```
DanielGuramulta, line 609: missing break again

apps/server/server.cpp
```diff
@@ -511,6 +505,155 @@ void invoke_sdk_api(payload::ClientRequest buff_recv) {
...
+        break;
```
DanielGuramulta, line 645: shouldn't this `break` be in the `if (!eeprom)` body? It always breaks and this line will never be reached ` 647: aditof::Status status = eeprom->close();`. Is this some temporary desired behaviour?

apps/uvc-app/uvc-gadget.cpp
```diff
@@ -1896,26 +1967,47 @@ int main(int argc, char *argv[]) {
...
     }
+    const aditof::DeviceConstructionData &camSensorData =
+        availableSensors[sensorIndex];
+    std::string availableSensorsBlob;
+    // Serialize information about sensors to be send to the UVC client
+    for (const auto &eeprom : camSensorData.eeproms) {
+        availableSensorsBlob += "EEPROM_NAME=" + eeprom.driverName + ";";
+        availableSensorsBlob += "EEPROM_PATH=" + eeprom.driverPath + ";";
+    }
+    uint32_t buffLen = availableSensorsBlob.length();
```
DanielGuramulta, line 1985: if `buffLen` is a read only variable, maybe make it `const`. From my point of view it increases the readability of the code. "oh, this is const, buffLen is a read only variable....";

sdk/src/connections/mipi/local_eeprom.cpp
```diff
@@ -0,0 +1,117 @@
...
+    if (ret < bytesCount) {
+        LOG(ERROR) << "fwrite() failed. Error: " << strerror(errno);
+        return Status::GENERIC_ERROR;
+    }
+    return Status::OK;
+}
+
+Status LocalEeprom::close() {
+    fclose(m_implData->eepromDev.fd);
+    m_implData->eepromDev.fd = NULL;
```
DanielGuramulta, line 108: `nullptr` instead of NULL?

sdk/src/connections/usb/linux/device_enumerator_linux.cpp
```diff
@@ -29,19 +29,53 @@
...
 #include <glog/logging.h>
 #include <linux/videodev2.h>
 #include <sys/stat.h>
 
-#define CLEAR(x) memset(&(x), 0, sizeof(x))
+using namespace std;
+
+static aditof::Status getAvailableSensors(int fd,
+                                          string &advertisedSensorData) {
+    int ret;
```
DanielGuramulta, line 47: Separate declaration and assignment, maybe `int ret = ...`

sdk/src/connections/usb/linux/device_enumerator_linux.cpp
```diff
@@ -29,19 +29,53 @@
...
+        fd, 4, 0, reinterpret_cast<uint8_t *>(&bufferLength),
+        sizeof(bufferLength));
+    if (ret < 0) {
+        LOG(WARNING)
+            << "Failed to read size of buffer holding sensors info. Error: "
+            << ret;
+        return aditof::Status::GENERIC_ERROR;
+    }
+
+    uint8_t *data = new uint8_t[bufferLength + 1];
```
DanielGuramulta, line 60: you could use `unique_ptr<uint8_t[]>`, no need to `delete[]` in two separate places afterwards. (pass .get() to the uvcExUnitReadBuffer)

sdk/src/connections/usb/linux/usb_linux_utils.h
```diff
@@ -0,0 +1,88 @@
+/*
```
DanielGuramulta, line 1: 👍

DanielGuramulta (approved):
Looks great! Will try to build and test this.

sdk/src/cameras/ad-96tof1-ebz/camera_96tof1.cpp
```diff
@@ -90,12 +95,30 @@ aditof::Status Camera96Tof1::initialize() {
     }
 
     // Initialize EEPROM
+    auto iter = std::find_if(m_devData.eeproms.begin(), m_devData.eeproms.end(),
+                             [](const EepromConstructionData &eData) {
+                                 return eData.driverName == "24c1024";
```
VLitanADI, line 100: `EEPROM_NAME` is already defined as `"24c1024"` in `target_definitions.h`. It would be more readable to use that define instead of the hard-coded value.

sdk/src/cameras/ad-96tof1-ebz/calibration_96tof1.cpp
```diff
@@ -113,8 +114,7 @@ aditof::Status Calibration96Tof1::displayCalMap() const {
 SaveCalMap - Saves the entire calibration map as binary to a file.
 \device - Pointer to a device instance
```
VLitanADI, line 115: Should be updated in order to match the new implementation (ex: `\eeprom - Reference to an eeprom instance`)

sdk/src/connections/mipi/dragonboard/target_definitions.h
```diff
@@ -32,6 +32,7 @@
 #ifndef TARGET_DEFINITIONS_H
 #define TARGET_DEFINITIONS_H
 
+static const char *EEPROM_NAME = "24c1024";
```
VLitanADI, line 35: This define is duplicated in each `target_definitions.h` of each platform. Maybe we should consider creating a `common_target_definitions.h` header as this definition is not platform-dependent.

VLitanADI (approved):

apps/server/server.cpp
```diff
@@ -213,6 +216,21 @@ int Network::callback_function(struct lws *wsi,
 
 void sigint_handler(int) { interrupted = 1; }
 
+shared_ptr<aditof::EepromInterface> findEepromByName(const string &name) {
```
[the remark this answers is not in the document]
dNechita: I switched to using std::

apps/server/server.cpp
dNechita: Refactored

apps/server/server.cpp
dNechita: fixed

apps/server/server.cpp
dNechita: fixed

apps/server/server.cpp
dNechita: yes, it should be inside the body. It's fixed now.

apps/uvc-app/uvc-gadget.cpp
dNechita: fixed

sdk/src/connections/mipi/local_eeprom.cpp
dNechita: yes, fixed

sdk/src/connections/usb/linux/device_enumerator_linux.cpp
dNechita: You mean to declare and assign in a single line (not to separate). Fixed.

sdk/src/connections/mipi/local_eeprom.cpp
```diff
@@ -29,65 +29,78 @@
...
 
-int eeprom_close(eeprom *e) {
-    if (e) {
-        fclose(e->fd);
-        e->fd = NULL;
-        e->valid = 0;
-    }
-    return 0;
+Status LocalEeprom::close() {
+    fclose(m_implData->eepromDev.fd);
```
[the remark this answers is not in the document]
dNechita: fixed

sdk/src/connections/mipi/local_eeprom.cpp
```diff
@@ -29,65 +29,78 @@
...
-    if (ret < size) {
-        fprintf(stderr, "eeprom_read_buf failed with %d: %s\n", errno,
-                strerror(errno));
-        return -1;
+Status LocalEeprom::read(const uint32_t address, uint8_t *data,
+                         const size_t bytesCount) {
+    auto fd = m_implData->eepromDev.fd;
+
+    fseek(fd, address, SEEK_SET);
+    size_t ret = fread(data, 1, bytesCount, fd);
```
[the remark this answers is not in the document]
dNechita: yes, they should be validated. Fixed.

sdk/src/connections/mipi/local_eeprom.cpp
```diff
@@ -29,65 +29,78 @@
...
-    if (ret < size) {
-        fprintf(stderr, "eeprom_write_buf failed with %d: %s\n", errno,
-                strerror(errno));
-        return -1;
+Status LocalEeprom::write(const uint32_t address, const uint8_t *data,
+                          const size_t bytesCount) {
+    auto fd = m_implData->eepromDev.fd;
+
+    fseek(fd, address, SEEK_SET);
+    size_t ret = fwrite(data, 1, bytesCount, fd);
```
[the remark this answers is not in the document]
dNechita: fixed

sdk/src/cameras/ad-96tof1-ebz/camera_96tof1.cpp
dNechita: target_definitions.h is part of the implementation of the low level API. The low level API doesn't need to expose the internals. The camera knows what EEPROM it is using. The best I can do is to declare a variable that holds the EEPROM name.

sdk/src/cameras/ad-96tof1-ebz/calibration_96tof1.cpp
dNechita: fixed

sdk/src/connections/mipi/dragonboard/target_definitions.h
dNechita: I would keep this as it is. This way, we have everything related to one target in the same place. If the list of duplicated things largely increases then yes, it would make sense to move them to some common header.

PR #400, opened 2020-09-10
apps/uvc-gadget: Fix writing to EEPROM

### LINKED ISSUE #395: EEPROM write error on Windows & Linux
The eeprom write fails on windows, and most probably on Linux too, when writing  the full firmware.

dNechita:
This should fix #395

### DIFF
--- a/sdk/src/connections/usb/windows/usb_eeprom_windows.cpp
+++ b/sdk/src/connections/usb/windows/usb_eeprom_windows.cpp
@@ -32,8 +32,10 @@
 #include "usb_eeprom.h"
 #include "usb_windows_utils.h"
 
+#include <chrono>
 #include <glog/logging.h>
 #include <string>
+#include <thread>
 
 using namespace aditof;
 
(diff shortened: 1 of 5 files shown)


sdk/src/connections/usb/windows/usb_eeprom_windows.cpp
```diff
@@ -101,6 +103,42 @@ Status UsbEeprom::write(const uint32_t address, const uint8_t *data,
...
+        --attempts;
+        if (FAILED(hr)) {
+            LOG(WARNING)
+                << "Failed to read a property via UVC extension unit. Error: "
+                << hr;
+            return Status::GENERIC_ERROR;
+        }
+
+        if (!eepromWriteStatus) {
+            std::this_thread::sleep_for(std::chrono::milliseconds(100));
```
DanielGuramulta, line 132: The SDK might be used in an application with a lot of threads (parallel image processing). Doing a busy wait (sleep) here would just block the current thread. Maybe use `std::this_thread::yield()`, allowing other threads to run. https://en.cppreference.com/w/cpp/thread/yield

dNechita: I will give it a try with this. One concern is that if there are very few other threads, the time checking and yielding will load the CPU while the sleep will not use CPU.

dNechita: Also, writing the EEPROM is something that is usually done during the initialization of the board and not while capturing and processing frames. In this case, using yield() will not do any good.

dNechita:
I consider this PR ready (no more implementation to add).
For this fix to work, one needs to recompile the kernel modules (using the changes in the patch included in this PR). Otherwise nothing will work, the camera won't even show up on the USB. Because of this I think this should be merged near to the moment we do another SD card image.

Issue #411, opened 2020-09-29
Python script to generate firmware file


acozma:
Create a Python script to generate a binary firmware file from all the .lf files in a folder.
The script must parse all the folders in the given folder, create the binaries for each folder and then combine them in a file which is the final eeprom image.

PR #412, opened 2020-09-30
Firmware generation tool

VLitanADI:
implements #411


tools/calibration-96tof1/cal_eeprom/cal_eeprom.py
```diff
@@ -34,471 +34,53 @@
...
-import sys
-from natsort import natsorted, ns
-import re
-import os
-import pandas as pd
-from . import firmware_gen as lf
-import json 
-import aditofpython as tof
-import tof_calib.device as device
+# import os
```
dNechita, line 40: Is this still required?

tools/calibration-96tof1/cal_eeprom/cal_map.py
```diff
@@ -0,0 +1,373 @@
...
+
+                sub_packet_value = struct.unpack('<f', r_b[j:j+4])
+                j = j+4
+                sub_packet_value = int(sub_packet_value[0])
+                i = i + 1
+                parameter_size = sub_packet_value
+                #print("Param Size", parameter_size)
+
+                number_of_elements = int(parameter_size/4)  # 4:size of float
+                #print("Number of elements", number_of_elements)
```
dNechita, line 363: Are all these prints required to be here?

dNechita (approved):
Looks good.

One thing that would be nice is to add a short description at the beginning of each script that explains the purpose of the script (what it can do and why it is useful).

tools/calibration-96tof1/cal_eeprom/cal_eeprom.py
VLitanADI: fixed.

tools/calibration-96tof1/cal_eeprom/cal_map.py
VLitanADI: not sure. I turned them into debug logs as they might be useful

PR #413, opened 2020-10-01
Prevent usage of invalid intrinsic parameters

VLitanADI:
Signed-off-by: Virgil Litan <virgil.litan@analog.com>

### DIFF
--- a/sdk/src/cameras/ad-96tof1-ebz/calibration_96tof1.cpp
+++ b/sdk/src/cameras/ad-96tof1-ebz/calibration_96tof1.cpp
@@ -478,7 +478,7 @@ void Calibration96Tof1::buildGeometryCalibrationCache(
     float x0 = cameraMatrix[2];
     float y0 = cameraMatrix[5];
 
-    bool validParameters = (fx != 0 && fy != 0);
+    const bool validParameters = (fx != 0 && fy != 0);
 
     if (m_geometry_cache) {
         delete[] m_geometry_cache;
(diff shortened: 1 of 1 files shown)


sdk/src/cameras/ad-96tof1-ebz/calibration_96tof1.cpp
```diff
@@ -490,7 +497,9 @@ void Calibration96Tof1::buildGeometryCalibrationCache(
             double tanYAngle = (y0 - i) / fy;
```
dNechita, line 497: Division by 0 will continue to exist. Would be nice to avoid that as well.

sdk/src/cameras/ad-96tof1-ebz/calibration_96tof1.cpp
```diff
@@ -478,10 +478,17 @@ void Calibration96Tof1::buildGeometryCalibrationCache(
     float x0 = cameraMatrix[2];
     float y0 = cameraMatrix[5];
 
+    bool validParameters = (fx != 0 && fy != 0);
```
DanielGuramulta, line 481: This can be const

VLitanADI: fixed.

sdk/src/cameras/ad-96tof1-ebz/calibration_96tof1.cpp
VLitanADI: fixed

dNechita (approved):

PR #424, opened 2020-11-11
GCC 7.3.0 fix for usb_linux_utils.cpp

murilopontes:
usb_linux_utils.cpp fix int ret initialization on line 134, this cause error on Yocto GCC 7.3.0


dNechita (approved):
Thanks!

PR #441, opened 2020-12-10
Extend frame details

VLitanADI:
Add `fullDataWidth` and `fullDataHeight` to `FrameDetails` and refactor the whole code (except uvc-gadget) in order to use the proper size depending on the needed values.
By this change there should be no more division / multiplication by 2 for frame size. Those operations were needed because on some platforms one frame is actually a container for 2 frames (IR and Depth).

### DIFF
--- a/examples/aditof-demo/aditofdemorecorder.cpp
+++ b/examples/aditof-demo/aditofdemorecorder.cpp
@@ -35,7 +35,7 @@
 #include <string.h>
 
 AditofDemoRecorder::AditofDemoRecorder()
-    : m_frameDetails{0, 0, ""}, m_recordTreadStop(true),
+    : m_frameDetails{0, 0, 0, 0, ""}, m_recordTreadStop(true),
       m_playbackThreadStop(true), m_shouldReadNewFrame(true),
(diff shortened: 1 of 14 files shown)


DanielGuramulta (commented):
It seems that in many cases the `fullDataWidth * fullDataHeight` are used in pointer arithmetics, or to express the size that the `Frame` object holds in memory, why not use something like `bytesUsed`. In some cases when playing with `uint16_t` pointers you will still get the `...bytesUsed / 2` but I think this is less confusing than having: `fullDataWidth`, `fullDataHeight` along with `width`, `height`

VLitanADI:
@DanielGuramulta Indeed this would be nicer but there is one place where the current solution is useful. The size of a frame received via USB will be `fullDataWidth` x `fullDataHeight`. More precisely here:
https://github.com/analogdevicesinc/aditof_sdk/blob/de9f110f2cab2928db3580434d08f2fe366086f5/sdk/src/connections/usb/linux/usb_device_linux.cpp#L198

DanielGuramulta (approved):

dNechita:
Shouldn't the network implementation be modified as well?

PR #460, opened 2020-12-18
Add multiple video dev support 

VLitanADI:
Adds support for xavier boards and fixes bugs on raspberry and jetson


examples/imshow-xavier/readme.md
```diff
@@ -0,0 +1,27 @@
+# Imshow jetson example
```
dNechita, line 1: 'jetson' word needs to be replaced with 'xavier', right?

VLitanADI: fixed

dNechita (approved):

PR #479, opened 2021-01-06
sdk: Fix overflow of data when writing it to the USB packet

dNechita:
A USB pachet has 60 bytes. First two contain medata, leaving 58 bytes for the payload.
But current implementation was writing 60 bytes (meaning 15 address-value pairs) over those 58 bytes.
This fixes things. If a address-value pair doesn't fit, then send address only in the first pachet and the send the value in the next pachet.

### DIFF
--- a/sdk/src/connections/usb/linux/usb_depth_sensor_linux.cpp
+++ b/sdk/src/connections/usb/linux/usb_depth_sensor_linux.cpp
@@ -485,17 +485,35 @@ aditof::Status UsbDepthSensor::writeAfeRegisters(const uint16_t *address,
     cq.size = MAX_BUF_SIZE;
 
     size_t sampleCnt = 0;
+    uint8_t b = sizeof(uint16_t); // Size (in bytes) of an AFE register
 
     length *= 2 * sizeof(uint16_t);
+
+    const uint8_t *pAddr = reinterpret_cast<const uint8_t *>(address);
+    const uint8_t *pData = reinterpret_cast<const uint8_t *>(data);
+    const uint8_t *ptr = pAddr;
+    bool pointingAtAddr = true;
+
     while (length) {
         memset(buf, 0, MAX_BUF_SIZE);
         buf[0] = length > MAX_PACKET_SIZE ? 0x01 : 0x02;
         buf[1] = length > MAX_PACKET_SIZE ? MAX_PACKET_SIZE : length;
-        for (int i = 0; i < buf[1]; i += 4) {
-            *(uint16_t *)(buf + 2 + i) = address[sampleCnt];
-            *(uint16_t *)(buf + 4 + i) = data[sampleCnt];
-            sampleCnt++;
+
+        for (int n = 0; n < buf[1]; ++n) {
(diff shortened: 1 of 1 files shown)


sdk/src/connections/usb/linux/usb_depth_sensor_linux.cpp
```diff
@@ -485,17 +485,35 @@ aditof::Status UsbDepthSensor::writeAfeRegisters(const uint16_t *address,
     cq.size = MAX_BUF_SIZE;
 
     size_t sampleCnt = 0;
+    uint8_t b = sizeof(uint16_t); // Size (in bytes) of an AFE register
```
VLitanADI, line 488: The name should be more suggestive. Maybe 'registerSize', 'regSize' etc. Also, this can be declared as 'const'. 
In order to be more rigorous I think 'sizeof(data[0])' should be used (but this is just a detail as 'sizeof(uint16_t)' is used in lots of other places, so this change would not make any difference)

sdk/src/connections/usb/linux/usb_depth_sensor_linux.cpp
```diff
@@ -485,17 +485,35 @@ aditof::Status UsbDepthSensor::writeAfeRegisters(const uint16_t *address,
...
         memset(buf, 0, MAX_BUF_SIZE);
         buf[0] = length > MAX_PACKET_SIZE ? 0x01 : 0x02;
         buf[1] = length > MAX_PACKET_SIZE ? MAX_PACKET_SIZE : length;
-        for (int i = 0; i < buf[1]; i += 4) {
-            *(uint16_t *)(buf + 2 + i) = address[sampleCnt];
-            *(uint16_t *)(buf + 4 + i) = data[sampleCnt];
-            sampleCnt++;
+
+        for (int n = 0; n < buf[1]; ++n) {
+            if ((sampleCnt / b) && (sampleCnt % b == 0)) {
```
VLitanADI, line 503: A short comment describing what happens here would be useful.
Maybe something like "once every two bytes the `ptr` is switch to point to the other array (data or addr) in order to interleave the values"

sdk/src/connections/usb/linux/usb_depth_sensor_linux.cpp
```diff
@@ -485,17 +485,35 @@ aditof::Status UsbDepthSensor::writeAfeRegisters(const uint16_t *address,
...
     while (length) {
         memset(buf, 0, MAX_BUF_SIZE);
         buf[0] = length > MAX_PACKET_SIZE ? 0x01 : 0x02;
         buf[1] = length > MAX_PACKET_SIZE ? MAX_PACKET_SIZE : length;
-        for (int i = 0; i < buf[1]; i += 4) {
-            *(uint16_t *)(buf + 2 + i) = address[sampleCnt];
-            *(uint16_t *)(buf + 4 + i) = data[sampleCnt];
-            sampleCnt++;
+
+        for (int n = 0; n < buf[1]; ++n) {
```
VLitanADI, line 502: `n` is conventionally used to denote the number of elements, I would suggest that we stick with the traditional `i` for loop iterators

PR #481, opened 2021-01-07
Update files to match FXTOF1 and RPI configuration

SeptimiuVana:


sdk/src/connections/target/addi9036_sensor.cpp
```diff
@@ -349,9 +349,11 @@ Addi9036Sensor::setFrameType(const aditof::FrameDetails &details) {
     /* Set the frame format in the driver */
     CLEAR(fmt);
     fmt.type = m_implData->videoBuffersType;
-#if defined FXTOF1
+
+    if (TARGET == "toybrick") {
```
dNechita, line 353: This will make compiler throw a warning "comparison with string literal results in unspecified behavior". To compare two C strings in C or C++ use strncmp().

sdk/src/connections/target/addi9036_sensor.cpp
```diff
@@ -349,9 +350,11 @@ Addi9036Sensor::setFrameType(const aditof::FrameDetails &details) {
     /* Set the frame format in the driver */
     CLEAR(fmt);
     fmt.type = m_implData->videoBuffersType;
-#if defined FXTOF1
+
+    if (!strncmp(TARGET, "toybrick", 4)) {
```
dNechita, line 354: Why 4?

sdk/src/connections/target/addi9036_sensor.cpp
dNechita: Also, while working on something else I noticed that this is not the right approach. The sdk/CMakeLists.txt no longer defines TARGET to "toybrick" but instead adds a new define: TOYBRICK. So you should use that.

PR #484, opened 2021-01-07
Fix documentation links

VLitanADI:


dNechita:
This branch also includes the changes from #485 which is also an active PR

dNechita (approved):

PR #493, opened 2021-01-11
Fixes for nx

VLitanADI:
Fixes: 
- Fix buffer size allocation for frame
- Replace deprecated `device` usage with `depthSensor`
- Naming convention fix method argument

### DIFF
--- a/sdk/src/frame_impl.cpp
+++ b/sdk/src/frame_impl.cpp
@@ -120,7 +120,7 @@ aditof::Status FrameImpl::getData(aditof::FrameDataType dataType,
 }
 
 void FrameImpl::allocFrameData(const aditof::FrameDetails &details) {
-    m_fullData = new uint16_t[details.fullDataWidth * details.fullDataHeight];
+    m_fullData = new uint16_t[details.width * details.height * 2];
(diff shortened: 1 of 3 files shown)


sdk/src/frame_impl.cpp
```diff
@@ -120,7 +120,7 @@ aditof::Status FrameImpl::getData(aditof::FrameDataType dataType,
 }
 
 void FrameImpl::allocFrameData(const aditof::FrameDetails &details) {
-    m_fullData = new uint16_t[details.fullDataWidth * details.fullDataHeight];
+    m_fullData = new uint16_t[details.width * details.height * 2];
```
dNechita, line 123: why multiply by 2? Isn't why fullDataWidth and height are there for?

VLitanADI: `fullDataHeight` and `fullDataWidth` describe the size of the full frame received from the camera. In the case of NX and AGX this is the same value as `width` and `height` because there are 2 virtual channels.
A frame at this level will always contain both IR and Depth, so it will always be twice the size of such a frame.

dNechita (approved):

Issue #461, opened 2020-12-21
Restructure temperature sensors


dNechita:
Rename Chicony -> TMP10X
Merge C implementation and C++ implementation for ADT7410 (and rename i2c_sensor -> ADT7410)

PR #496, opened 2021-01-14
Sensors restructure

dNechita:
Addresses #461, #462


SeptimiuVana (approved):

sdk/src/cameras/ad-fxtof1-ebz/camera_fxtof1.cpp
```diff
@@ -62,10 +62,10 @@ static const std::vector<std::string> availableControls = {
 static const std::string skEepromName = "custom";
 CameraFxTof1::CameraFxTof1(
     std::shared_ptr<aditof::DepthSensorInterface> depthSensor,
-    std::shared_ptr<aditof::StorageInterface> eeprom,
-    std::shared_ptr<aditof::TemperatureSensorInterface> temperatureSensor)
-    : m_depthSensor(depthSensor), m_eeprom(eeprom),
-      m_temperatureSensor(temperatureSensor), 
+    std::vector<std::shared_ptr<aditof::StorageInterface>> &eeprom,
+    std::vector<std::shared_ptr<aditof::TemperatureSensorInterface>> &tSensor)
+    : m_depthSensor(depthSensor), m_eeprom(eeprom.first()),
```
VLitanADI, line 67: What happens if the vector is empty?
Are we sure that this will never happen?

sdk/src/connections/target/addi9036_sensor.cpp
```diff
@@ -394,11 +375,9 @@ Addi9036Sensor::setFrameType(const aditof::FrameDetails &details) {
         /* Set the frame format in the driver */
         CLEAR(fmt);
         fmt.type = dev->videoBuffersType;
-        
-#if defined TOYBRICK
+        //#if defined TOYBRICK //????
```
VLitanADI, line 378: The following line was causing an error on other boards. 
Will test again.

sdk/src/connections/target/addi9036_sensor.cpp
```diff
@@ -86,10 +72,11 @@ struct VideoDev {
 };
 
 struct Addi9036Sensor::ImplData {
+    uint8_t numVideoDevs;
```
VLitanADI, line 75: this should be const.

sdk/src/connections/target/adt7410_sensor.cpp
```diff
@@ -29,33 +29,96 @@
...
+
+struct ADT7410::ImplData {
+    temp_sensor tDev;
+    std::string driverPath;
+    int i2c_address;
+};
+
+ADT7410::ADT7410(const std::string &name, const std::string &driver_path,
+                 int i2c_address)
+    : m_implData(new ImplData), m_name(name) {
```
VLitanADI, line 56: Why not use initializer list for all members?

sdk/src/cameras/ad-fxtof1-ebz/camera_fxtof1.cpp
dNechita: Actually, I forgot to rework the FXTOF1 camera. This won't even compile.

sdk/src/connections/target/addi9036_sensor.cpp
dNechita: @acozma  was thinking that it might work. We should test, yes.

sdk/src/connections/target/adt7410_sensor.cpp
dNechita: Which member isn't initialized?

VLitanADI: this is actually ok, my bad :)

sdk/src/connections/target/addi9036_sensor.cpp
dNechita: It can't be. By default it is set to 1 when ImplData is constructed but then a value gets assigned to it in the open() function of Addi9036Sensor.

sdk/src/connections/target/addi9036_sensor.cpp
VLitanADI: turns out it does not. will add the macro back

dNechita: ok. On which platform it did not work?

PR #503, opened 2021-01-18
Add new script to change build files depending on configuration

SeptimiuVana:

### DIFF
--- a/scripts/build_camera.sh
+++ b/scripts/build_camera.sh
@@ -1,0 +1,19 @@
+#!/bin/bash
+
+cd $(dirname $(dirname $(realpath $0)))
+git pull
+sudo rm -r build
+mkdir build
+cd build
+
+if [ "$#" -eq 2 ]; then 
+	cmake "$1" "$2" -DWITH_OPENCV=on -DWITH_PYTHON=on -DCMAKE_PREFIX_PATH="/opt/glog;/opt/protobuf;/opt/websockets" .. 
+
+elif [ "$#" -eq 1 ]; then
+	cmake "$1" -DWITH_OPENCV=on -DWITH_PYTHON=on -DCMAKE_PREFIX_PATH="/opt/glog;/opt/protobuf;/opt/websockets" .. 
+
+else
+	cmake -DWITH_OPENCV=on -DWITH_PYTHON=on -DCMAKE_PREFIX_PATH="/opt/glog;/opt/protobuf;/opt/websockets" .. 	
+fi
+
+sudo cmake --build . --target install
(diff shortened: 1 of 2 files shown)


scripts/build_camera.sh
```diff
@@ -0,0 +1,19 @@
+#!/bin/bash
+
+cd ..
```
VLitanADI, line 3: Using this assumes that the script is run from inside the folder `scripts` which is not guaranteed.

VLitanADI: one way to make sure that you go into the root of the project is to use 
`cd $(dirname $(dirname $(realpath $0)))`
this assumes that the script is placed one level below the root, but it can be ran from wherever.

scripts/build_camera.sh
```diff
@@ -0,0 +1,19 @@
+#!/bin/bash
```
dNechita, line 1: Please add a description of what arguments this script expects

dNechita:
This script should be called: rebuild_for_camera.sh. And an explanation/description of why and when this is needed should be added.

VLitanADI:
if no argument is provided, it would be nice to print a help message.

scripts/rebuild_for_camera.sh
```diff
@@ -0,0 +1,27 @@
+#This script can be used to rebuild the SDK depending on various platforms and cameras.
+#Its purpose is to be called from a shortcut placed on desktop.
+#It expects the parameters passed as the CMAKE command.
+
+#Possible command for the camera: -DUSE_FXTOF1=1 
+#Possible commands for the board: -DDRAGONBOARD=1 -DRASPBERRYPI=1 -DJETSON=1 -DXAVIERNX=1 -DXAVIER=1
+#The implicit build, if no parameters are passed is for AD-96TOF1-EBZ build on linux.
+
+#!/bin/bash
```
VLitanADI, line 9: The shebang should be on the first line of the script in order to have any effect.
https://stackoverflow.com/questions/12910744/why-should-the-shebang-line-always-be-the-first-line

scripts/rebuild_for_camera.sh
```diff
@@ -0,0 +1,27 @@
...
+
+cd $(dirname $(dirname $(realpath $0)))
+git pull
+sudo rm -r build
+mkdir build
```
VLitanADI, line 14: use `mkdir -p build` in order to avoid error messages if the folder already exists.

scripts/rebuild_for_camera.sh
```diff
@@ -0,0 +1,27 @@
+#This script can be used to rebuild the SDK depending on various platforms and cameras.
+#Its purpose is to be called from a shortcut placed on desktop.
+#It expects the parameters passed as the CMAKE command.
+
+#Possible command for the camera: -DUSE_FXTOF1=1
```
VLitanADI, line 5: `command` is not a good word for this. Maybe `build flag` or `argument` would be more appropriate.

dNechita (approved):

PR #517, opened 2021-02-03
fix format with clang-format6.0

VLitanADI:
Signed-off-by: Virgil Litan <virgil.litan@analog.com>


format.sh
```diff
@@ -0,0 +1,48 @@
+is_source_file() {
```
dNechita, line 1: Can you add at least a small description of what the script does. What if formats, if it takes arguments, and what files will ignore? Just to have a general idea of why this script exist.

DanielGuramulta (approved):
Looks good. I think we should not format cvui.h as it is a file not maintained by us. If we will ever update it in the future we might have conflicts by styling it. It can be added to the .clangformatignore file

Issue #466, opened 2020-12-22
Implement compatibility check


VLitanADI:
Implement a mechanism that allows checking the sdk version and compatibility. Useful when not running on target and different versions may be used

VLitanADI:
- The compatibility check shall be done at the level of uvc-gadget / server. 
- In a dedicated header for each module the server, uvc-gadget and api shall have defined the Major, and Minor (the api already has this header sdk/include/version.h)
- The compatibility shall be done by comparing the string sent by the remote with the following string:

<API_MAJOR>_<API_MINOR>_<UVC_MAJOR>_<UVC_MINOR> or <API_MAJOR>_<API_MINOR>_<SERVER_MAJOR>_<SERVER_MINOR>

- When the string match the connection should continue as usual, otherwise an error is returned, containing the expected string.
- When receiving a special string (ex "IGNORE_TARGET_VERSION") the connection shall be accepted
- If the connection fails the remote shall print the expected string and the one it tried to use.

Implementation suggestions: 

- The comparison should be done in a dedicated class in order to allow easier development for new comparison rules. Also, this allows the server and uvc gadget to call the same piece of code
- use sdk/include/version.h as an example for uvc / server version headers

PR #520, opened 2021-02-05
Add version check at connection

VLitanADI:
implements issue #466


apps/server/server.cpp
```diff
@@ -620,6 +621,12 @@ void invoke_sdk_api(payload::ClientRequest buff_recv) {
         break;
     }
```
dNechita, line 623: Why do you need the server to tell what connection type it is? It can’t be anything else than Network.

VLitanADI:
TODO after discussion with @dNechita :

- [ ] use hacky include instead of exposing `connection_validator` to the clients of the sdk
- [ ] rename `getConnectionString` to something more suggestive: `getConnectionToken`, `getVersionString` ?

apps/server/server.cpp
dNechita: Ignore this. I misunderstood things.

sdk/src/connections/network/network_depth_sensor.cpp
```diff
@@ -85,6 +86,7 @@ NetworkDepthSensor::~NetworkDepthSensor() {
 aditof::Status NetworkDepthSensor::open() {
     using namespace aditof;
 
+    std::string connectionString;
```
dNechita, line 89: This doesn't seem to be used anywhere, nor the connection_validator.h.

VLitanADI: removed it.

sdk/src/connections/network/network_sensor_enumerator.cpp
```diff
@@ -44,18 +45,57 @@ NetworkSensorEnumerator::NetworkSensorEnumerator(const std::string &ip)
...
 
     if (net->ServerConnect(m_ip) != 0) {
         LOG(WARNING) << "Server Connect Failed";
         return Status::UNREACHABLE;
     }
 
+    getVersionString(net, connectionString);
+
+    if (!isValidConnection(aditof::ConnectionType::NETWORK, connectionString)) {
+        LOG(ERROR) << "invalid connection string: " << connectionString;
```
dNechita, line 95: now, it's version string. Update err message and variable name

dNechita (approved):

dNechita:
You need to fix build on Windows before merging!

PR #530, opened 2021-02-18
sdk: camera: xavier agx: Update subdev name for L4T 32.5

btogorean:
In L4T 32.5 kernel update the video subdevs enumeration order was
changed. Update SDK accordingly.

Signed-off-by: btogorean <bogdan.togorean@analog.com>


dNechita:
Wouldn't this break things for those who didn't get the kernel update?

btogorean:
> Wouldn't this break things for those who didn't get the kernel update?

you're right but what's on master now is

PR #539, opened 2021-03-08
sdk: Add 'attributes' to a frame

dNechita:
Different cameras can stream different types of frames. Attributes can
be specific to a type of frame. Attributes are identified by name (string)
and have a value (also string).

This allows for keeping the same API for frames that are somehow
different but not fundamentally different.

### DIFF
--- a/sdk/include/aditof/frame.h
+++ b/sdk/include/aditof/frame.h
@@ -93,12 +93,47 @@ class SDK_API Frame {
     Status getDetails(FrameDetails &details) const;
 
     /**
+     * @brief Gets details of a type of data within the frame
+     * @param dataType
+     * @param[out] details
+     * @return Status
+     */
+    Status getDataDetails(const std::string &dataType,
+                          FrameDataDetails &details) const;
+
+    /**
      * @brief Gets the address where the specified data is being stored
      * @param dataType
      * @param[out] dataPtr
      * @return Status
      */
-    Status getData(FrameDataType dataType, uint16_t **dataPtr);
+    Status getData(const std::string &dataType, uint16_t **dataPtr);
+
+    /**
+     * @brief Gets the frames's list of attributes
+     * @param[out] attributes
+     * @return Status
+     */
+    virtual Status
+    getAvailableAttributes(std::vector<std::string> &attributes) const = 0;
+
+    /**
+     * @brief Sets a specific frame attribute
+     * @param[in] attribute - Attribute name
+     * @param[in] value - Attribute value
+     * @return Status
+     */
+    virtual Status setAttribute(const std::string &attribute,
+                                const std::string &value) = 0;
+
+    /**
+     * @brief Gets the value of a specific frame attribute
+     * @param[in] attribute - Attribute name
+     * @param[out] value - Attribute value
+     * @return Status
+     */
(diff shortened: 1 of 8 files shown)


sdk/src/cameras/chicony-006/camera_chicony_006.cpp
```diff
@@ -336,7 +336,7 @@ aditof::Status CameraChicony::requestFrame(aditof::Frame *frame,
     }
 
     uint16_t *frameDataLocation;
-    frame->getData(FrameDataType::FULL_DATA, &frameDataLocation);
+    frame->getData(allData, &frameDataLocation);
```
VLitanADI, line 339: missing `" "`

sdk/src/frame_impl.cpp
```diff
@@ -97,30 +93,36 @@ aditof::Status FrameImpl::getDetails(aditof::FrameDetails &details) const {
...
-    m_irData = m_fullData + (details.width * details.height);
+    unsigned int totalWidth = 0;
+    unsigned int totalHeigth = 0;
+
+    for (const auto &details : details.dataDetails) {
+        totalWidth += details.width;
+        totalHeigth += details.height;
+    }
+
+    m_allData = new uint16_t[totalWidth * totalHeigth];
```
VLitanADI, line 120: why not use `details.width` and `details.height`?

sdk/include/aditof/frame_definitions.h
```diff
@@ -55,30 +68,31 @@ enum class FrameDataType {
...
      */
-    unsigned int height;
+    std::vector<FrameDataDetails> dataDetails;
 
     /**
-     * @brief The width of the actual full frame.
+     * @brief The width of the entire frame
      */
-    unsigned int fullDataWidth;
+    unsigned int width;
```
VLitanADI, line 85: Given that the sizes of the contained frames are specified in `dataDetails` I think it would be more useful to have just `size` in `FrameDetails`. This would make more sense in the case we include metadata in the frame (which is not organized in a matrix fashion)

sdk/src/frame_impl.cpp
dNechita: Actually this is where details.width and details.height should be calculated. However, I don't think we need them at all (outside FrameImpl).
I don't think we will be adding metadata for this camera. For other cameras, they will have their own FrameImpl.

sdk/include/aditof/frame.h
```diff
@@ -92,13 +92,48 @@ class SDK_API Frame {
...
+                          FrameDataDetails &details) const;
+
     /**
      * @brief Gets the address where the specified data is being stored
      * @param dataType
      * @param[out] dataPtr
      * @return Status
      */
-    Status getData(FrameDataType dataType, uint16_t **dataPtr);
+    Status getData(const std::string &dataType, uint16_t **dataPtr);
```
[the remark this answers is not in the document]
VLitanADI: As long as it does not modify the contents of the class it can be declared as `const`. I don`t think that the the contents of the memory where the pointers point are members of the class.

dNechita:
Closing this as it has been moved to aditof-sdk-rework.

PR #553, opened 2021-04-21
Fix geometry calibration cache

acozma:


dNechita (approved):

dNechita:
The commit contained by this PR is already in master (it's a different commit but with the same changes). So I think we can close this PR. Is this ok?

acozma:
The change on master is in a different file, this one is for the ros bindings.
https://github.com/analogdevicesinc/aditof_sdk/blob/master/sdk/src/cameras/ad-fxtof1-ebz/calibration_fxtof1.cpp#L395

dNechita:
Ok

PR #559, opened 2021-05-13
Add different parameter for depth correction (similar to 96tof1 camera)

stefansobol:


stefansobol:
Instead of using "camera_geometry_correction" to disable/enable depth calibration and camera geometry calibration, use a different control ("depth_correction") for depth calibration. Similar implementation as for [camera 96tof1](https://github.com/analogdevicesinc/aditof_sdk/blob/b6205fbe416c0cf6f36cd976e584e0081e700520/sdk/src/cameras/ad-96tof1-ebz/camera_96tof1.cpp#L447).

acozma (approved):

PR #570, opened 2021-07-23
Update aditof demo to support setFrameType function

SeptimiuVana:


examples/first-frame-network/main.cpp
```diff
@@ -73,7 +73,7 @@ int main(int argc, char *argv[]) {
         std::cout << "no frame type available!";
         return 0;
     }
-    status = camera->setFrameType(frameTypes.front());
+    status = camera->setFrameType("depth_only");
```
dNechita, line 76: I think this change was just to test things, right?

SeptimiuVana: If we keep it this way we can use the example with jetson as well, it can be reverted.

sdk/src/cameras/ad-fxtof1-ebz/camera_fxtof1.cpp
```diff
@@ -188,13 +189,15 @@ aditof::Status CameraFxTof1::initialize() {
 }
 
 aditof::Status CameraFxTof1::start() {
-    // return m_depthSensor->start(); // For now we keep the device open all the time
-    return aditof::Status::OK;
+    return m_depthSensor
+        ->start(); // For now we keep the device open all the time
```
dNechita, line 193: The comments no longer make sense if you uncommented this line.

sdk/src/cameras/ad-fxtof1-ebz/camera_fxtof1.cpp
```diff
@@ -308,12 +323,18 @@ aditof::Status CameraFxTof1::setFrameType(const std::string &frameType) {
     }
 
     if (m_details.frameType != *frameDetailsIt) {
+        uint16_t afeRegsAddr[2] = {0x4001, 0x7c22};
+        uint16_t afeRegsVal[2] = {0x0006, 0x0004};
+        m_depthSensor->writeAfeRegisters(afeRegsAddr, afeRegsVal, 2);
```
dNechita, line 328: What is happening here? Why is this required?

SeptimiuVana: We need to turn off the streaming before setting the frame type and then turning it back on.

dNechita: This needs to be documented next to the code.

examples/first-frame-network/main.cpp
dNechita: Also, if we keep this, it won't work for remote<->USB<->target setup because in this scenario there is only "depth_ir" (https://github.com/analogdevicesinc/aditof_sdk/blob/master/sdk/src/connections/usb/linux/usb_depth_sensor_linux.cpp#L176 and https://github.com/analogdevicesinc/aditof_sdk/blob/master/sdk/src/connections/usb/windows/usb_depth_sensor_windows.cpp#L523)

dNechita: you could set the frame type with the first element from frameTypes. This way it's more generic.

SeptimiuVana: Fixed.

sdk/src/cameras/ad-fxtof1-ebz/camera_fxtof1.cpp
SeptimiuVana: Fixed.

sdk/src/cameras/ad-fxtof1-ebz/camera_fxtof1.cpp
SeptimiuVana: Fixed.

sdk/src/cameras/ad-fxtof1-ebz/camera_fxtof1.cpp
```diff
@@ -332,6 +329,7 @@ aditof::Status CameraFxTof1::setFrameType(const std::string &frameType) {
             return status;
         }
         m_details.frameType = *frameDetailsIt;
+        //Turn of the streaming of data
```
dNechita, line 332: off (typo) or should it be on?

sdk/src/connections/target/addi9036_sensor.cpp
```diff
@@ -369,6 +369,16 @@ Addi9036Sensor::setFrameType(const aditof::FrameDetails &details) {
...
+            CLEAR(req);
+	        req.count = 0;
+	        req.type = dev->videoBuffersType;
+	        req.memory = V4L2_MEMORY_MMAP;
+
+	        if (xioctl(dev->fd, VIDIOC_REQBUFS, &req) == -1) {
+	            LOG(WARNING) << "VIDIOC_REQBUFS error "
+	                         << "errno: " << errno << " error: " << strerror(errno);
+	            return Status::GENERIC_ERROR;
+
```
dNechita: Why is this needed?

sdk/src/cameras/ad-fxtof1-ebz/camera_fxtof1.cpp
SeptimiuVana: It should be on. Sorry, i will fix this.

sdk/src/connections/target/addi9036_sensor.cpp
SeptimiuVana: This was added to clear the buffers from the driver.

dNechita: ok. Can you add a short comment with this information?

SeptimiuVana: Fixed.

dNechita (approved):

PR #572, opened 2021-07-30
server/server.cpp: fix variable name

SeptimiuVana:


dNechita (approved):

dNechita:
It would be nice if users could provide the port (if not,  a default would be used as it happens today) at the same moment when they provide the IP.

PR #590, opened 2021-09-08
examples/aditof-demo/aditofdemoview.cpp: enable the input of all letters

SeptimiuVana:

### DIFF
--- a/examples/aditof-demo/aditofdemoview.cpp
+++ b/examples/aditof-demo/aditofdemoview.cpp
@@ -41,48 +41,17 @@
 
 namespace detail {
 std::string getKeyPressed(int key, bool &backspace) {
-    switch (key & 0xff) {
-    case 46:
-        return ".";
-    case 48:
-        return "0";
-    case 49:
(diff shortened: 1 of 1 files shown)


examples/aditof-demo/aditofdemoview.cpp
```diff
@@ -41,48 +41,17 @@
...
-
-    default:
+    key = key & 0xff;
+    if (key < 128) {
+        if (key == 8) {
+            backspace = true;
+            return "";
+        }
+        char c = key;
+        std::string val(&c);
```
dNechita, line 51: what if dot character is being entered? Will it work?

dNechita: I suppose it will. But please make sure to test if it works.

SeptimiuVana: With the latest update it works. It has been tested on linux, windows and dragonboard.

dNechita (approved):

PR #592, opened 2021-09-09
Low-level api for rgb camera

rbudai98:


sdk/src/connections/target/rgb_cam.h
```diff
@@ -0,0 +1,108 @@
+/*
```
dNechita, line 1: rename file to rgb_sensor to match the class name

dNechita:
Those 9 commits created by me shouldn't be here. They already exist in master. Please remove them.

sdk/src/cameras/3d-smart-camera/camera_3d_smart.cpp
```diff
@@ -84,6 +86,9 @@ Camera3D_Smart::Camera3D_Smart(
         LOG(WARNING) << "Invalid instance of a RGB sensor";
         return;
     }
+    //aditof::SensorDetails sDetails;
+    //m_rgbSensor->getDetails(sDetails);
+    //m_details.connection = sDetails.connectionType;
```
dNechita, line 91: At this point we already know the connection type from the other sensor.

sdk/src/cameras/3d-smart-camera/camera_3d_smart.cpp
```diff
@@ -366,6 +378,7 @@ aditof::Status Camera3D_Smart::setFrameType(const std::string &frameType) {
         uint16_t afeRegsVal[2] = {0x0006, 0x0004};
         m_depthSensor->writeAfeRegisters(afeRegsAddr, afeRegsVal, 2);
         status = m_depthSensor->setFrameType(*frameDetailsIt);
+        status = m_rgbSensor->setFrameType(*frameDetailsIt);
```
dNechita, line 381: rgbSensor can't be set with the same frame type that is also used to set the depth sensor.

dNechita: This PR should focus on the RGBSensor class as described by the task and not on the Camera3DSmart class which is part of the high level API. I know that some stuff are missing from the camera class but that should be handled in a separate task/PR.

sdk/src/connections/target/addi9036_sensor.cpp
```diff
@@ -609,9 +609,9 @@ aditof::Status Addi9036Sensor::getFrame(uint16_t *buffer,
     } else if (!isBufferPacked(buf[0], width, height)) {
         // TODO: investigate optimizations for this (arm neon / 1024 bytes
         // chunks)
-        if (m_implData->frameDetails.type == "depth_only") {
+        if (m_implData->frameDetails.type.find("depth_only") != std::string::npos) {
```
dNechita, line 612: These shouldn't be changed.

sdk/src/frame_impl.cpp
```diff
@@ -118,7 +109,7 @@ aditof::Status FrameImpl::getData(aditof::FrameDataType dataType,
 
     switch (dataType) {
     case FrameDataType::FULL_DATA: {
-        *dataPtr = m_fullData;
+        *dataPtr = m_fullData;
```
dNechita, line 112: Remove trailing spaces. Use format tool.

dNechita (changes requested):

sdk/src/connections/target/rgb_sensor.cpp
```diff
@@ -0,0 +1,663 @@
...
+
+        if (close(dev->fd) == -1) {
+            LOG(WARNING) << "close m_implData->fd error "
+                         << "errno: " << errno << " error: " << strerror(errno);
+        }
+
+        /*if (close(dev->sfd) == -1) {
+            LOG(WARNING) << "close m_implData->sfd error "
+                         << "errno: " << errno << " error: " << strerror(errno);
+        }*/
```
dNechita, line 134: If this is not needed, delete it.

sdk/src/connections/target/rgb_sensor.cpp
```diff
@@ -0,0 +1,663 @@
...
+                }
+            }
+        } else {
+            /* Error ignored */
+        }
+
+        /* Set the frame format in the driver */
+        CLEAR(fmt);
+        fmt.type = V4L2_BUF_TYPE_VIDEO_CAPTURE;
+        fmt.fmt.pix.width = 1920;
```
dNechita, line 236: Why not use the macros defined in header?

sdk/src/connections/target/rgb_sensor.cpp
```diff
@@ -0,0 +1,663 @@
...
+            return Status::GENERIC_ERROR;
+    }
+    return status;
+}
+
+aditof::Status
+RgbSensor::getAvailableFrameTypes(std::vector<aditof::FrameDetails> &types) {
+    using namespace aditof;
+    Status status = Status::OK;
+    for (int i = 0; i < types.size(); i++) {
```
dNechita, line 318: types is an output parameter. This means that this function will set a value to it and not rely on it's old value. You don't need a for here. There is only one type that you need to add which is called: "rgb".

sdk/src/connections/target/rgb_sensor.cpp
```diff
@@ -0,0 +1,663 @@
...
+                return Status::GENERIC_ERROR;
+        }
+    }
+
+    m_implData->frameDetails = details;
+
+    return status;
+}
+
+aditof::Status RgbSensor::program(const uint8_t *firmware, size_t size) {
```
dNechita, line 403: Is RgbSensor programable?

sdk/src/connections/target/rgb_sensor.cpp
```diff
@@ -0,0 +1,663 @@
...
+        }
+        unsigned int buf_data_len;
+        uint8_t *pdata[m_implData->numVideoDevs];
+        status = getInternalBufferPrivate(&pdata[i], buf_data_len, buf[i], dev);
+        if (status != Status::OK) {
+            return status;
+        }
+
+        int k = 0;
+        for (unsigned int j = 0; j < (buf_data_len); j = j + 2) {
```
dNechita, line 510: no need for round brackets around buf_data_len

sdk/src/connections/target/rgb_sensor.cpp
```diff
@@ -0,0 +1,663 @@
...
+            buffer[k++] = (pdata[i][j + 1] << 8) + pdata[i][j];
+        }
+        //memcpy(buffer, pdata[i], sizeof(buffer));
```
dNechita, line 513: If not needed, remove it please

sdk/src/connections/target/rgb_sensor.cpp
```diff
@@ -0,0 +1,663 @@
...
+
+        status = enqueueInternalBufferPrivate(buf[i], dev);
+        if (status != Status::OK) {
+            return status;
+        }
+    }
+    bufferInfo->timestamp =
+        buf[0].timestamp.tv_sec * 1000000 + buf[0].timestamp.tv_usec;
+    //memcpy(buffer, pdata[0], sizeof(pdata[0]));
```
dNechita, line 522: remove if not needed

sdk/src/connections/target/rgb_sensor.cpp
```diff
@@ -0,0 +1,663 @@
...
+    return status;
+}
+
+aditof::Status RgbSensor::readAfeRegisters(const uint16_t *address,
+                                           uint16_t *data, size_t length) {
+    using namespace aditof;
+    Status status = Status::OK;
+    return status;
```
dNechita, line 530: We should return the UNAVAILABLE error. If this sensor doesn't support this operation then it's best to say so.

sdk/src/cameras/3d-smart-camera/jetson/sensor_enumerator_jetson.cpp
```diff
@@ -0,0 +1,145 @@
...
+    dev_name = str;
+    subdev_name = str;
+    return Status::OK;
+}
+
+Status TargetSensorEnumerator::searchSensors() {
+    Status status = Status::OK;
+    LOG(INFO) << "Looking for devices on the target: Jetson";
+
+    // TO DO: Don't guess the device, find a way to identify it so we are sure
```
dNechita, line 90: This TO DO message is no longer needed. We are now identifying the sensors and no longer guessing their paths.

sdk/src/cameras/3d-smart-camera/camera_3d_smart.cpp
```diff
@@ -334,17 +335,23 @@ aditof::Status Camera3D_Smart::getAvailableModes(
 aditof::Status Camera3D_Smart::setFrameType(const std::string &frameType) {
     using namespace aditof;
     Status status = Status::OK;
+    //int poz = frameType.find('-');
+    //std::string leftFrameType = frameType.substr(0, poz);
+    //std::string rightFrameType = frameType.substr(poz+1, frameType.size()-poz-1);
```
dNechita, line 340: please remove these lines

sdk/src/cameras/3d-smart-camera/camera_3d_smart.cpp
```diff
@@ -334,17 +335,23 @@ aditof::Status Camera3D_Smart::getAvailableModes(
...
 
     if (m_devStarted) {
         status = m_depthSensor->stop();
+        status = m_rgbSensor->stop();
         if (status != Status::OK) {
             return status;
         }
         m_devStarted = false;
     }
+    //depth_only, ir_only and depth_ir
```
dNechita, line 350: this comment doesn't seem relevant, please remove

sdk/src/connections/target/rgb_sensor.cpp
rbudai98: As for the last week's discussion I was told that rgb camera shouldn't have a separate frame type, rather try to combine the already existing types with rgb, since we use them in parrallel. The following options were given: 

- depth_only-rgb
- ir_only-rgb
- depth_ir-rgb
 in order to combine the naming I just had to add to the already existing frametypes the new naming

sdk/src/connections/target/rgb_sensor.cpp
rbudai98: nope, deleted

sdk/src/connections/target/rgb_sensor.cpp
rbudai98: done

sdk/src/connections/target/rgb_sensor.cpp
rbudai98: done

sdk/src/connections/target/rgb_sensor.cpp
rbudai98: done

sdk/src/connections/target/rgb_sensor.cpp
rbudai98: modified, with every other non-relevant functions included

sdk/src/connections/target/rgb_sensor.cpp
rbudai98: done

sdk/src/connections/target/rgb_sensor.cpp
dNechita: Things discussed are still true. But the combined frame type belong to the camera. These changes are in rgb sensor not in the camera. The sensor currently has one type and that is rgb.

dNechita (approved):

PR #596, opened 2021-09-13
Add rgb view in aditof-demo

SeptimiuVana:


examples/aditof-demo/aditofdemoview.h
```diff
@@ -91,6 +89,12 @@ class AdiTofDemoView {
     bool localModeChecked = false;
     bool ethModeChecked = false;
     int connectionCurrentValue = 4;
+
+    bool depthIrChecked = true;
+    bool depthOnlyChecked = false;
+    bool irOnlyChecked = false;
+    int frameTypeCurrentValue = 4;
+    // 4 = depthIr(default); 2 = depthOnly; 1 = irOnly
```
dNechita, line 101: can you put this comment on the same line as the line code?

dNechita:
There is a lot of code that is enabled only if smart 3D camera is enabled. Wouldn't it be possible to analyze the frame type of the camera and if it contains RBG content then do all the call to be able to display the RGB image?

SeptimiuVana:
> There is a lot of code that is enabled only if smart 3D camera is enabled. Wouldn't it be possible to analyze the frame type of the camera and if it contains RBG content then do all the call to be able to display the RGB image?

We could but it will add a few more conditions to be verified if the 3D camera is disabled.

examples/aditof-demo/aditofdemoview.h
dNechita: yes, that might be a little better

dNechita:
> > There is a lot of code that is enabled only if smart 3D camera is enabled. Wouldn't it be possible to analyze the frame type of the camera and if it contains RBG content then do all the call to be able to display the RGB image?
> 
> We could but it will add a few more conditions to be verified if the 3D camera is disabled.
That is not an impediment or is it?

SeptimiuVana:
> > > There is a lot of code that is enabled only if smart 3D camera is enabled. Wouldn't it be possible to analyze the frame type of the camera and if it contains RBG content then do all the call to be able to display the RGB image?
> > 
> > 
> > We could but it will add a few more conditions to be verified if the 3D camera is disabled.
> 
> That is not an impediment or is it?

It will do extra checks for the other systems in which we do not have the rgb camera and it might affect the speed.

dNechita:
> > > > There is a lot of code that is enabled only if smart 3D camera is enabled. Wouldn't it be possible to analyze the frame type of the camera and if it contains RBG content then do all the call to be able to display the RGB image?
> > > 
> > > 
> > > We could but it will add a few more conditions to be verified if the 3D camera is disabled.
> > 
> > 
> > That is not an impediment or is it?
> 
> It will do extra checks for the other systems in which we do not have the rgb camera and it might affect the speed.

Aditof-demo is an example that uses the SDK. The SDK exposes a generic API for a camera (it abstract over different types of cameras). This abstraction is one of the core features. If you use preprocessor directives to tell the demo which camera is using then what's the point in having a generic API? If we can't do it without directives then the API needs improving.
The demo should be able to ask the camera what are its capabilities and construct an UI according to that.
I don't think a few IFs should prevent us from doing this.

SeptimiuVana:
> > > > > There is a lot of code that is enabled only if smart 3D camera is enabled. Wouldn't it be possible to analyze the frame type of the camera and if it contains RBG content then do all the call to be able to display the RGB image?
> > > > 
> > > > 
> > > > We could but it will add a few more conditions to be verified if the 3D camera is disabled.
> > > 
> > > 
> > > That is not an impediment or is it?
> > 
> > 
> > It will do extra checks for the other systems in which we do not have the rgb camera and it might affect the speed.
> 
> Aditof-demo is an example that uses the SDK. The SDK exposes a generic API for a camera (it abstract over different types of cameras). This abstraction is one of the core features. If you use preprocessor directives to tell the demo which camera is using then what's the point in having a generic API? If we can't do it without directives then the API needs improving.
> The demo should be able to ask the camera what are its capabilities and construct an UI according to that.
> I don't think a few IFs should prevent us from doing this.

Very good point, I will modify the code.

examples/aditof-demo/aditofdemoview.h
SeptimiuVana: fixed

sdk/src/frame_impl.cpp
```diff
@@ -130,8 +130,9 @@ aditof::Status FrameImpl::getData(aditof::FrameDataType dataType,
 }
 
 void FrameImpl::allocFrameData(const aditof::FrameDetails &details) {
-    m_fullData = new uint16_t[details.width * details.height * 2 +
-                              details.rgbWidth * details.rgbHeight];
+    //removed the usage of rgbWidth/height to avoid app crash when running on target
```
dNechita, line 133: What is this about?

dNechita (approved):
Rest looks good

sdk/src/frame_impl.cpp
SeptimiuVana: Fixed. Found another way to disable the allocation

PR #597, opened 2021-09-14
sdk: Implement getFrameTimestamp for network/target depth sensor

SeptimiuVana:


sdk/include/aditof/depth_sensor_interface.h
```diff
@@ -104,6 +104,14 @@ class DepthSensorInterface {
      */
     virtual aditof::Status getFrame(uint16_t *buffer) = 0;
 
+    /**
+     * @brief Request timestamp of the requested frame
+     * @param timestamp - a valid location where the new timestamp should be stored.
+     * @return Status
+     */
+
+    virtual aditof::Status getFrameTimestamp(long long &timestamp) = 0;
```
dNechita, line 113: I would do things differently. I would modify the getFrame() so that it retrieves the timestamp as well (besides the buffer). This way we get all we need (buffer, timestamp) in a single transaction. I would even make the second parameter a structure (let's say bufferInfo declared maybe in sensor_definitions.h) which would hold the timestamp and could hold other info in the future.
It's not wrong having getFrameTimestamp() method but it adds confusion. It raises questions like: for which frame does it return the timestamp? is it the last one? what if I call this before getFrame()? what if I close the sensor, can I still get the timestamp? what if the first call to getFrame() fails (maybe a timeout) and the second call works fine, does it get the right timestamp(for the second frame)? What if I call getFrame() and in a separate thread I call getFrameTimestamp()?
Any thoughts? @SeptimiuVana @acozma

SeptimiuVana: Well we will get the timestamp of the last frame ( if it was successfully requested) because it is set when we request a frame. The problem might appear (as you mentioned above) when we try to get a timestamp if we do not request a frame.(It will throw an error with no timestamp). 
It should be better if we get both the frame and the timestamp in getFrame() but we can add the timestamp as an default parameter.

sdk/include/aditof/camera.h
```diff
@@ -123,6 +123,7 @@ class SDK_API Camera {
      * @return Status
      */
     virtual Status requestFrame(Frame *frame,
+                                aditof::BufferInfo *bufferInfo = nullptr,
```
dNechita, line 126: BufferInfo is a low level API structure. Doesn't make sense to show up in the high level API.

dNechita: To get the timestamp into the Frame object, we should create a timestamp field for it and copy the value from the BufferInfo object. And this should happen in the camera implementation.

sdk/include/aditof/camera_definitions.h
```diff
@@ -34,6 +34,7 @@
 
 #include "connections.h"
 #include "frame_definitions.h"
+#include "sensor_definitions.h"
```
dNechita, line 37: Won't be needing this.

sdk/include/aditof/depth_sensor_interface.h
```diff
@@ -113,6 +115,7 @@ class DepthSensorInterface {
      * @param length - the number of registers to read
      * @return Status
      */
+
```
dNechita, line 118: is this a product of clang-format?

sdk/include/aditof/sensor_definitions.h
```diff
@@ -57,6 +57,13 @@ struct SensorDetails {
     ConnectionType connectionType;
 };
 
+struct BufferInfo {
+    /**
+     * @brief Timestamp of the last requested frame
```
dNechita, line 62: "Timestamp of the data within the buffer" should be more appropriate

sdk/src/connections/target/addi9036_sensor.cpp
```diff
@@ -538,7 +538,7 @@ aditof::Status Addi9036Sensor::program(const uint8_t *firmware, size_t size) {
     return status;
 }
 
-aditof::Status Addi9036Sensor::getFrame(uint16_t *buffer) {
+aditof::Status Addi9036Sensor::getFrame(uint16_t *buffer, aditof::) {
```
dNechita, line 541: this is missing the variable type and name

sdk/src/cameras/ad-fxtof1-ebz/camera_fxtof1.cpp
```diff
@@ -381,12 +382,12 @@ aditof::Status CameraFxTof1::requestFrame(aditof::Frame *frame,
...
     frame->getData(FrameDataType::FULL_DATA, &frameDataLocation);
 
-    status = m_depthSensor->getFrame(frameDataLocation);
+    status = m_depthSensor->getFrame(frameDataLocation, bufferInfo);
     if (status != Status::OK) {
         LOG(WARNING) << "Failed to get frame from device";
         return status;
     }
-
+    LOG(INFO) << "TIMESTAMP: " << bufferInfo->timestamp;
```
dNechita, line 390: It would be too noisy to print timestamp for each frame while streaming a video.

sdk/include/aditof/camera.h
SeptimiuVana: Removed bufferInfo from high level API

sdk/include/aditof/camera_definitions.h
SeptimiuVana: Fixed.

sdk/include/aditof/depth_sensor_interface.h
SeptimiuVana: No, fixed.

sdk/include/aditof/sensor_definitions.h
SeptimiuVana: Fixed.

sdk/src/cameras/ad-fxtof1-ebz/camera_fxtof1.cpp
SeptimiuVana: It is just for test purposes, will be removed.

sdk/src/cameras/ad-fxtof1-ebz/camera_fxtof1.cpp
```diff
@@ -381,7 +382,8 @@ aditof::Status CameraFxTof1::requestFrame(aditof::Frame *frame,
     uint16_t *frameDataLocation;
     frame->getData(FrameDataType::FULL_DATA, &frameDataLocation);
 
-    status = m_depthSensor->getFrame(frameDataLocation);
+    aditof::BufferInfo bufferInfo;
+    status = m_depthSensor->getFrame(frameDataLocation, &bufferInfo);
```
dNechita, line 386: We're working on adding the 3D Smart Camera. This should have been added there and not in FXTOF1.

sdk/src/connections/network/network_depth_sensor.cpp
```diff
@@ -401,6 +402,9 @@ aditof::Status NetworkDepthSensor::getFrame(uint16_t *buffer) {
                              m_implData->frameDetails_cache.fullDataHeight);
     }
 
+    if (bufferInfo != nullptr) {
```
dNechita, line 405: This can be simplified: if (bufferInfo)

sdk/src/connections/target/addi9036_sensor.h
```diff
@@ -103,6 +104,7 @@ class Addi9036Sensor : public aditof::DepthSensorInterface,
     std::string m_driverSubPath;
     std::string m_captureDev;
     std::unique_ptr<ImplData> m_implData;
+    long long timestamp;
```
dNechita, line 107: You don't really need this. You can obtain the timestamp right in the getFrame() implementation.

sdk/src/connections/usb/linux/usb_depth_sensor_linux.cpp
```diff
@@ -32,6 +32,7 @@
 #include "connections/usb/usb_depth_sensor.h"
 #include "usb_linux_utils.h"
 #include "utils.h"
+#include <aditof/sensor_definitions.h>
```
dNechita, line 35: is this needed?

sdk/src/cameras/ad-fxtof1-ebz/camera_fxtof1.cpp
```diff
@@ -37,6 +37,7 @@
 #include "sensor_names.h"
 #include <aditof/frame.h>
 #include <aditof/frame_operations.h>
+#include <aditof/sensor_definitions.h>
```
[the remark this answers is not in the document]
SeptimiuVana: Fixed.

sdk/src/cameras/ad-fxtof1-ebz/camera_fxtof1.cpp
SeptimiuVana: Implemented in all cameras.

sdk/src/connections/target/addi9036_sensor.h
SeptimiuVana: Yes, indeed. Fixed

sdk/src/connections/usb/linux/usb_depth_sensor_linux.cpp
SeptimiuVana: No, fixed.

sdk/src/connections/network/network_depth_sensor.cpp
SeptimiuVana: yes, indeed, thanks!

dNechita (approved):

PR #605, opened 2021-09-27
apps/server/server.cpp: extend server to use rgb sensor

SeptimiuVana:

### DIFF
--- a/apps/server/buffer.proto
+++ b/apps/server/buffer.proto
@@ -58,6 +58,7 @@ message ClientRequest
   repeated bytes func_bytes_param = 40; // List of function parameters of type byte-array
   bool expect_reply = 50;               // Whether a response with data is expected or not
   FrameDetails frame_type = 60;         // Frame type information
+  int32 id = 70;
 }
 
 message ServerResponse
@@ -72,4 +73,5 @@ message ServerResponse
   SensorsInfo sensors_info = 80;                     // List of information about existing devices on the same platform as the server
   string message = 90;                               // Additional message (if any)
   BufferInfo buffer_details = 100;                   // Buffer details
+  int32 id = 110;
 }
--- a/apps/server/buffer.proto
+++ b/apps/server/buffer.proto
@@ -40,7 +40,7 @@ message SensorsInfo
 {
   repeated StorageInfo storages = 10;               // Information about available storages
   repeated TemperatureSensorInfo temp_sensors = 20; // Information about available temperature sensors
-  ImagerInfo image_sensors = 30;                    // Information about available image sensors
+  repeated ImagerInfo image_sensors = 30;                    // Information about available image sensors
 }
 
 message BufferInfo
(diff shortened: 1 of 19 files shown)


apps/server/server.cpp
```diff
@@ -314,13 +320,16 @@ void invoke_sdk_api(payload::ClientRequest buff_recv) {
...
-        aditof::SensorDetails depthSensorDetails;
-        camDepthSensor->getDetails(depthSensorDetails);
+
+        int depth_sensor_id = 0;
         auto pbSensorsInfo = buff_send.mutable_sensors_info();
-        sensorV4lBufAccess =
-            std::dynamic_pointer_cast<aditof::V4lBufferAccessInterface>(
-                camDepthSensor);
+        for (const auto &depthSensor : depthSensors) {
+            camDepthSensor[depth_sensor_id] = depthSensor;
```
dNechita, line 327: depth_sensor_id should be incremented in this loop. Otherwise it will always be 0.

SeptimiuVana: yes, thank you! fixed

apps/server/buffer.proto
```diff
@@ -58,6 +58,7 @@ message ClientRequest
   repeated bytes func_bytes_param = 40; // List of function parameters of type byte-array
   bool expect_reply = 50;               // Whether a response with data is expected or not
   FrameDetails frame_type = 60;         // Frame type information
+  int32 id = 70;
```
dNechita, line 61: It would make more sense to follow the structure of the other sensors (temperature, storage).
I mean have another message called ImagerInfo with fields name and id. And add it to the SensorsInfo message.

sdk/include/aditof/sensor_definitions.h
```diff
@@ -45,12 +45,23 @@ namespace aditof {
  * @struct SensorDetails
  * @brief Provides details about the device
  */
-struct SensorDetails {
+struct SensorInformations {
```
dNechita, line 48: This seems to be used internally only. For this reason it shouldn't be public. Also you could have directly 2 members (id and name) instead of this struct.

apps/server/buffer.proto
```diff
@@ -34,6 +40,7 @@ message SensorsInfo
 {
   repeated StorageInfo storages = 10;               // Information about available storages
   repeated TemperatureSensorInfo temp_sensors = 20; // Information about available temperature sensors
+  repeated ImagerInfo image_sensors = 30;                    // Information about available image sensors
```
dNechita, line 43: would be nice to align the comment with the rest of comments

apps/server/buffer.proto
```diff
@@ -58,6 +65,7 @@ message ClientRequest
   repeated bytes func_bytes_param = 40; // List of function parameters of type byte-array
   bool expect_reply = 50;               // Whether a response with data is expected or not
   FrameDetails frame_type = 60;         // Frame type information
+  SensorsInfo sensors_info = 70;        // List of informations regarding which sensor to address
```
dNechita, line 68: maybe reword in a simpler way: "The sensor to address" or "The sensor for which the request should apply"

PR #609, opened 2021-09-29
rgb_sensor: added converter from bayer into rgb

rbudai98:


sdk/src/connections/target/rgb_sensor.h
```diff
@@ -103,6 +123,9 @@ class RgbSensor : public aditof::DepthSensorInterface,
     std::string m_driverSubPath;
     std::string m_captureDev;
     std::unique_ptr<ImplData> m_implData;
+    int Rw = 256; //scaling monitor: https://en.wikipedia.org/wiki/Color_balance
```
dNechita, line 126: Please use the same notation style as the rest of members (use m_ prefix)

sdk/src/connections/target/rgb_sensor.cpp
```diff
@@ -580,3 +579,157 @@ aditof::Status RgbSensor::getInternalBuffer(uint8_t **buffer,
...
+                rgb[(i * width + j) * 3 + GREEN] = (uint8_t)(
+                    ((float)directCopy(pData, i, j, width, height) / Rw) *
+                    255.0);
+                rgb[(i * width + j) * 3 + BLUE] = (uint8_t)(
+                    ((float)verticalKernel(pData, i, j, width, height) / Bw) *
+                    255.0);
+            }
+        }
+    }
+    buffer = (uint16_t *)rgb;
```
dNechita, line 734: this line has no effect. Buffer and rgb hold the same value.

sdk/src/connections/target/rgb_sensor.cpp
```diff
@@ -580,3 +579,157 @@ aditof::Status RgbSensor::getInternalBuffer(uint8_t **buffer,
...
+}
```
dNechita, line 735: missing new line at the end of file

sdk/src/connections/target/rgb_sensor.cpp
```diff
@@ -580,3 +579,157 @@ aditof::Status RgbSensor::getInternalBuffer(uint8_t **buffer,
...
+
+void RgbSensor::bayer2RGB(uint16_t *buffer, uint8_t *pData, int width,
+                          int height) {
+    //casting into 8 in order to work with it
+    uint8_t *rgb = (uint8_t *)buffer;
+    //uint16_t *rgb = buffer;
+
+    for (int i = 0; i < height; i++) {
+        for (int j = 0; j < width; j++) {
+            //rgb[(i*width+j)*RED] =
```
dNechita, line 681: is this needed?

sdk/src/connections/target/rgb_sensor.cpp
```diff
@@ -580,3 +579,157 @@ aditof::Status RgbSensor::getInternalBuffer(uint8_t **buffer,
...
+uint16_t RgbSensor::directCopy(uint8_t *pData, int x, int y, int width,
+                               int height) {
+    return getValueFromData(pData, x, y, width, height); //
+}
+
+void RgbSensor::bayer2RGB(uint16_t *buffer, uint8_t *pData, int width,
+                          int height) {
+    //casting into 8 in order to work with it
+    uint8_t *rgb = (uint8_t *)buffer;
+    //uint16_t *rgb = buffer;
```
dNechita, line 677: remove if not needed

sdk/src/connections/target/rgb_sensor.cpp
```diff
@@ -580,3 +579,157 @@ aditof::Status RgbSensor::getInternalBuffer(uint8_t **buffer,
...
+    }
+    if (x - 1 >= 0 && x - 1 < height && y + 1 >= 0 && y + 1 < width) {
+        sum += getValueFromData(pData, x - 1, y + 1, width, height);
+        nr++;
+    }
+    if (x + 1 >= 0 && x + 1 < height && y + 1 >= 0 && y + 1 < width) {
+        sum += getValueFromData(pData, x + 1, y + 1, width, height);
+        nr++;
+    }
+    return ((uint16_t)((float)sum / (float)nr));
```
dNechita, line 665: What if execution doesn't go through any of the IFs? Dividing 0 by 0 isn't desired.

dNechita: Also you could remove the round brackets at the edges. They don't do anything but make expression less readable.

sdk/src/connections/target/rgb_sensor.cpp
```diff
@@ -580,3 +579,157 @@ aditof::Status RgbSensor::getInternalBuffer(uint8_t **buffer,
 aditof::Status RgbSensor::enqueueInternalBuffer(struct v4l2_buffer &buf) {
     return enqueueInternalBufferPrivate(buf);
 }
+
+void RgbSensor::setRw(int value) { this->Rw = value; }
```
dNechita, line 583: having a notation with m_ prefix let's us know we're dealing with a member of this instance. So this isn't really necessary.

sdk/src/connections/target/rgb_sensor.cpp
```diff
@@ -432,10 +432,9 @@ aditof::Status RgbSensor::getFrame(uint16_t *buffer,
             return status;
         }
 
-        int k = 0;
-        for (unsigned int j = 0; j < buf_data_len; j = j + 2) {
-            buffer[k++] = (pdata[i][j + 1] << 8) + pdata[i][j];
-        }
+        //conversion done on the target
+        RgbSensor::bayer2RGB(buffer, pdata[i], RGB_FRAME_WIDTH,
```
dNechita, line 436: not a real problem but I don't think "RgbSensor::" adds value. We know bayerRGB is a method of this class. I don't think it helps explicitly saying so.

dNechita (approved):

PR #619, opened 2021-10-14
3d_cam: Frame synchronization

rbudai98:


dNechita:
Closing as we no longer need to synchronize frames.

PR #622, opened 2021-10-15
ros: rgb message topic 

rbudai98:


dNechita (approved):

bindings/ros/aditof_roscpp/src/examples/rviz_pointcloud.cpp
```diff
@@ -43,7 +43,7 @@ int main(int argc, char **argv) {
         return -1;
     }
 
-    setFrameType(camera, "depth_ir");
+    setFrameType(camera, "depth_ir_rgb");
```
SeptimiuVana, line 46: The other cameras will not work with this frame type.

PR #645, opened 2021-10-29
Fix crash when running on smart camera target

acozma:


dNechita (approved):
It slipped while solving conflicts.

PR #653, opened 2021-11-01
Sdk and server fixes

dNechita:


apps/server/server.h
```diff
@@ -33,7 +33,7 @@
 #include <google/protobuf/io/zero_copy_stream_impl.h>
 #include <libwebsockets.h>
 
-#define RX_BUFFER_BYTES (4147900)
+#define RX_BUFFER_BYTES (4147900 * 2)
```
SeptimiuVana, line 36: Why do we need so much bytes?

dNechita: The RGB image is: 4147200 bytes (1920 * 1080 * 2 bytes). The extra 700 are for additional things in the protobuf message (status, error string, etc.). But this doesn't cover the depth+ir so @acozma doubled the size. I can only assume this is why he did it. Maybe he can answer better.
The depth+ir size would be: 1228800 (640 * 960 * 2 bytes).

PR #688, opened 2021-11-15
Changed call "getImageSensor" to "getImageSensors"

tmanalog:
(in low_level_example.py)


dNechita:
Can you search through all pythons examples and see if they need fixing?
Also, can you check the calibration tool as well?

tmanalog:
Yes, I am going to look at all the examples in python. Where is the calibration tool?

dNechita (approved):

PR #689, opened 2021-11-15
sdk/cameras: read serial from eeprom and populate cameraId

SeptimiuVana:

### DIFF
--- a/sdk/src/cameras/3d-smart-camera/camera_3d_smart.cpp
+++ b/sdk/src/cameras/3d-smart-camera/camera_3d_smart.cpp
@@ -183,6 +183,20 @@ aditof::Status Camera3D_Smart::initialize() {
     }
     m_tempSensorsInitialized = true;
 
+    // Read the camera's serial number from eeprom
+    uint8_t *eepromSerial = new uint8_t(EEPROM_SERIAL_LENGHT);
+    status =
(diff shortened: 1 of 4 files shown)


sdk/src/cameras/3d-smart-camera/camera_3d_smart.cpp
```diff
@@ -183,6 +183,20 @@ aditof::Status Camera3D_Smart::initialize() {
     }
     m_tempSensorsInitialized = true;
 
+    // Read the camera's serial number from eeprom
+    uint8_t *eepromSerial = new uint8_t(EEPROM_SERIAL_LENGHT);
+    status =
+        m_eeprom->read(EEPROM_SERIAL_ADDR, eepromSerial, EEPROM_SERIAL_LENGHT);
+    if (status != Status::OK) {
```
dNechita, line 190: Memory leaks if status is not ok because you don't free the buffer allicated above. It's easier to declar the buffer on the stack, especially if size is small.

dNechita (changes requested):

sdk/src/cameras/3d-smart-camera/camera_3d_smart.cpp
SeptimiuVana: Good point, I will modify the code, thank you.

SeptimiuVana: fixed

dNechita (approved):

Issue #589, opened 2021-09-08
Implement setFrameType function for USB communication


SeptimiuVana:
Update setFrameType function in the following files:
*  usb_depth_sensor_linux
* usb_depth_sensor_windows
* camera_96tof

rbudai98:
#696 for camera_96tof

PR #696, opened 2021-11-16
Add96 setFrameType correction

rbudai98:


rbudai98:
Related issue: #589

dNechita (approved):
Tested on USB and seems that functionality is not affected (I was worried that stopping and starting the frame acquisition would not work due to these changes).

PR #697, opened 2021-11-16
added functionality for logging the SDK version, branch and commit

tmanalog:


sdk/src/system_impl.cpp
```diff
@@ -104,7 +104,11 @@ buildCamera(std::unique_ptr<SensorEnumeratorInterface> enumerator) {
     return camera;
 }
 
-SystemImpl::SystemImpl() {}
+SystemImpl::SystemImpl() {
+    LOG(INFO) << "SDK version: " << (std::string)VERSION
```
dNechita, line 108: Make sure this prints only once no matter how many instances of SystemImpl we create.

tmanalog:
Ok.

CMakeLists.txt
```diff
@@ -82,6 +82,40 @@ if (USE_3D_SMART)
...
+	if ("${ADITOFSDK_GIT_REPO}" STREQUAL "${CMAKE_CURRENT_SOURCE_DIR}")
+		execute_process(
+			COMMAND ${GIT_EXECUTABLE} rev-parse --short HEAD
+			WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}
+			OUTPUT_VARIABLE ADITOFSDK_GIT_COMMIT
+			OUTPUT_STRIP_TRAILING_WHITESPACE
+		)
+	endif()
+
+	if ("${ADITOFSDK_GIT_REPO}" STREQUAL "${CMAKE_CURRENT_SOURCE_DIR}")
```
dNechita, line 105: This IF is the same as the previous one so you can put both commands in one IF.

dNechita (approved):

PR #702, opened 2021-11-18
solved a few warnings: [-Wsign-compare] and [-Wreorder]

tmanalog:


examples/aditof-demo/aditofdemoview.h
```diff
@@ -61,36 +61,50 @@ class AdiTofDemoView {
   private:
     std::shared_ptr<AdiTofDemoController> m_ctrl;
     std::string m_viewName;
+
```
dNechita, line 64: Is is the order in the constructor (in the .cpp file) that is wrong not in the header.

PR #703, opened 2021-11-18
solved a few warnings: [-Wsign-compare] and [-Wreorder]

tmanalog:

### DIFF
--- a/sdk/src/cameras/3d-smart-camera/calibration_3d_smart.cpp
+++ b/sdk/src/cameras/3d-smart-camera/calibration_3d_smart.cpp
@@ -482,8 +482,8 @@ aditof::Status Calibration3D_Smart::distortionCorrection(uint16_t *frame,
             int x_dist = (int)(x_dist_adim * fx + cx);
             int y_dist = (int)(y_dist_adim * fy + cy);
 
-            if (x_dist >= 0 && x_dist < width && y_dist >= 0 &&
-                y_dist < height) {
+            if (x_dist >= 0 && x_dist < (int)width && y_dist >= 0 &&
+                y_dist < (int)height) {
                 buff[j * width + i] = frame[y_dist * width + x_dist];
             } else {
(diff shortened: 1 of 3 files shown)


sdk/src/cameras/3d-smart-camera/calibration_3d_smart.cpp
```diff
@@ -479,8 +479,8 @@ aditof::Status Calibration3D_Smart::distortionCorrection(uint16_t *frame,
             double y_dist_adim = y * m_distortion_cache[j * width + i];
 
             //back to original space
-            int x_dist = (int)(x_dist_adim * fx + cx);
-            int y_dist = (int)(y_dist_adim * fy + cy);
+            unsigned int x_dist = (unsigned int)(x_dist_adim * fx + cx);
```
dNechita, line 482: I wouldn't cast to unsigned here. If by any chance, the expression returns a negative value, casting it to an only positive value will populate x_dist and y_dist with undesired values. You can cast them inside the if (xdist >= 0 && x_dist < width &...)

sdk/src/cameras/3d-smart-camera/calibration_3d_smart.cpp
```diff
@@ -482,8 +482,8 @@ aditof::Status Calibration3D_Smart::distortionCorrection(uint16_t *frame,
             int x_dist = (int)(x_dist_adim * fx + cx);
             int y_dist = (int)(y_dist_adim * fy + cy);
 
-            if (x_dist >= 0 && x_dist < width && y_dist >= 0 &&
-                y_dist < height) {
+            if (x_dist >= 0 && (unsigned int)x_dist < width && y_dist >= 0 &&
```
dNechita, line 485: The problem is still there. You need to cast width to signed since we know the value won't be too high to overflow.

dNechita (approved):

PR #707, opened 2021-11-23
Changed the order of the embedded platforms

acozma:


dNechita:
Looks good. Additionally, we could change the link where "user guide" points to. It's the user guides in section: 'Supported Platforms'. Currently it points to 96TOF1 and should point to 3D Smart Camera user guide.

dNechita (approved):

PR #723, opened 2021-12-14
Dragonboard usb mode toggle bug fixed

rbudai98:
Signed-off-by: rbudai98 <robert.budai@analog.com>


sdk/src/cameras/ad-96tof1-ebz/camera_96tof1.cpp
```diff
@@ -217,11 +217,13 @@ aditof::Status Camera96Tof1::initialize() {
 }
 
 aditof::Status Camera96Tof1::start() {
+    m_devStarted = true;
     return m_depthSensor
         ->start(); // For now we keep the device open all the time
```
dNechita, line 222: Can you please remove the comment since it is in contradiction to what the code does?

dNechita (approved):

PR #730, opened 2021-12-16
examples: Include ImGui library only if WITH_IMGUI is enabled

dNechita:

### DIFF
--- a/examples/CMakeLists.txt
+++ b/examples/CMakeLists.txt
@@ -5,7 +5,9 @@ project(examples)
 option(WITH_IMGUI "Build ImGui based examples?" OFF)
 
 # Dependecies that can be accessed by any example
-add_subdirectory(deps/imgui)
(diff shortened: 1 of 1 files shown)


examples/CMakeLists.txt
```diff
@@ -5,7 +5,9 @@ project(examples)
 option(WITH_IMGUI "Build ImGui based examples?" OFF)
 
 # Dependecies that can be accessed by any example
-add_subdirectory(deps/imgui)
+if (WITH_IMGUI)
```
SeptimiuVana, line 8: Can we add this in the same if with aditof-demo-imgui?

dNechita: I would keep it here. I wanted to separate the dependencies that can be shared between the examples from the examples.

SeptimiuVana (approved):

PR #737, opened 2022-01-03
Create python binding for first frame network

tmanalog:


SeptimiuVana:
Please remove duplicated commits.

bindings/python/examples/first_frame_network/first_frame_network.py
```diff
@@ -0,0 +1,84 @@
...
+
+if len(sys.argv) < 2 :
+    print("No ip provided! Run: python3 first_frame_network.py ip!")
+    exit(1)
+
+ip = sys.argv[1]
+
+cameras = []
+status = system.getCameraListAtIp(cameras, ip)
+print("system.getCameraList()", status)
```
dNechita, line 46: Should be: getCameraListAtIp

bindings/python/examples/first_frame_network/readme.md
```diff
@@ -0,0 +1,65 @@
+# First frame Example
```
dNechita, line 1: This readme talks about too much code which is already done in the first_frampe.py example. It would be better to talk only about getCameraAtIp(). Also it would be nice to give an example of how it can be used (use a real ip).

dNechita (approved):

PR #756, opened 2022-02-09
XAVIER_NX: setting registers for enabling stream separated for specific

rbudai98:
channels

IR and DEPTH bug, was not working induvidually, fixed with this commit
Signed-off-by: rbudai <robert.budai@analog.com>


rbudai98:
Xaviern NX by using current device tree is not able to turn off ir or depth hence causing crash when selecting only one frame type.
In order to fix this error we keep the register setting always at depth-ir (register 0xc3da at value 0x0007) in order to obtain frames. Also framesize has to remain at 640x480x2 constant due to the same fact.

PR #761, opened 2022-02-22
Changed m_mutex from simple mutex to recursive_mutex

tmanalog:


dNechita:
Can you please describe why this change is needed?

tmanalog:
This change is needed because the mutex is acquired multiple times by the same thread. In order to support this functionality, the mutex must be recursive.

acozma (approved):

PR #770, opened 2022-03-24
Add support for VS 17 2022

tmanalog:


dNechita:
There is this task. Adding support for Visual Studio 2022 also means adding an additional job in the Azure CI.
Also, make sure to keep 2019 as well. The VS 2022 should not replace 2019.

dNechita:
Looks good. I am waiting for you to mention in the windows build instructions that we also support VS2022 and then this can be merged.

dNechita (approved):

PR #777, opened 2022-04-05
Macos Build

cristina-suteu:
fixed methods so the sdk can be built on macOS
added build instructions for macOS

### DIFF
--- a/sdk/src/connections/usb/macos/usb_depth_sensor_macos.cpp
+++ b/sdk/src/connections/usb/macos/usb_depth_sensor_macos.cpp
@@ -33,9 +33,6 @@
 
 struct UsbDepthSensor::ImplData {};
 
-//UsbDepthSensor::UsbDepthSensor(
-//    const aditof::DeviceConstructionData & /*data*/) {}
-//
 UsbDepthSensor::UsbDepthSensor(const std::string &driverPath)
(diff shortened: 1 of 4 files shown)


tmanalog (commented):
I think there should be a doxygen tutorial here too, in the build instructions.

sdk/src/connections/usb/macos/usb_depth_sensor_macos.cpp
```diff
@@ -33,8 +33,8 @@
 
 struct UsbDepthSensor::ImplData {};
 
-UsbDepthSensor::UsbDepthSensor(
-    const aditof::DeviceConstructionData & /*data*/) {}
+UsbDepthSensor::UsbDepthSensor(const std::string &driverPath)
+    : m_driverPath(driverPath), m_implData(new UsbDepthSensor::ImplData) {}
```
dNechita, line 37: Since you allocate memory for ImplData with new you must delete it at some point. E.g. in the destructor.

dNechita: actually, I've checked and m_implData is a unique_ptr so all is good. Sorry!

dNechita (approved):

PR #796, opened 2022-07-06
Convert sdk code from C++ to JavaScript

tmanalog:


tmanalog (commented):
I organized all the JS code we had, until now, in just one directory: bindings/javascript/.

PR #801, opened 2022-08-08
Xavier AGX device tree updated according to 5.3 kernel version

rbudai98:
Signed-off-by: rbudai98 <robert.budai@analog.com>


dNechita:
Do these changes affect other setups (AGX+96tof1)?

rbudai98:
> 
In this form only FXTOF1 is connected with the specific hardware modifications

dNechita (approved):

